PR #1, opened 2022-01-28
fixed clay electrolysis

dtcameron:
This reverts clay electrolysis back to LV as it was in NF. As the change from LV-> MV was not a CE -> CEu change.


m2r1k5:
I will adjust the recipe to be in line with the CEu recipe just accessible at LV, but this can be merged for now

PR #59, opened 2022-07-13
Add information to Quest #353

CrossElectronics:
The pack focuses more on GT tools, so I think this is a good time to introduce GT's butchery knife.


tracer4b:
This looks like the wrong questbook file. (In fact, I have no idea why and how this file is in this repository. It's the regular Nomi quest book lang file.)
But I will probably put this in the questbook next round I update it. Thank you

CrossElectronics:
No problem. But I actually tested myself before modifying on the repo. It is quite weird if that is not the correct file

PR #60, opened 2022-07-19
Add simplified chinese(zh_cn) translation (partially)

704009905:
Adds translation to zh_cn, mainly craft tweaking contents.

I'm new to gregtech and this modpack, most terminology translations are based on chemistry naming, gregtech's lang file and playtesting game context.

For zh_ch translation of mods used in this pack, I would suggest [this resource pack](https://github.com/CFPAOrg/Minecraft-Mod-Language-Package).

I'm also making a quest book translation, but nomi-ceu didn't use the language key feature in betterquesting mod. To cleanly support multi-language it's required to transform quest contents to language key. (I did this too but it conflicts with 1.3.5). Contentwise the quest book is translated 70% and checked 20%. My playthrough is just entering HV so many contents are not thoroughly tested.


tracer4b:
For the quest book, the current solution to swap between questbooks is honestly quite bad. I'll try to work something out so that both questbooks are compatible with language keys. Maybe you could PR the change you made to have language keys work on the hard mode qb, and we can work from there.

There are also some item tooltips and names hard-defined in CT scripts - those need to be moved into language keys as well.

As for this PR contents, I looked at the translations and they seemed fine, but my Chinese is not that good. For now this can be merged and anything with issues can be edited later. Thank you for your contribution 😀

PR #66, opened 2022-08-01
Changed "Fluid Transport" quest text

Terlen:
Changed quest text to:
1) provide a better example of increased pipe throughput vs EnderIO conduits
2) remove reference to incorrect transfer rate of Normal Potin Fluid Pipe.

Not sure if Small Potin Pipe retrieval task should be changed or removed? Really no reason to make it instead of the Pressurized Conduit early game.

### DIFF
--- a/overrides/config/betterquesting/DefaultQuests.json
+++ b/overrides/config/betterquesting/DefaultQuests.json
@@ -3470,7 +3470,7 @@
       "properties:10": {
         "betterquesting:10": {
           "autoclaim:1": 0,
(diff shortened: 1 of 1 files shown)


Exaxxion:
Might be worth specifying you're talking about Pressurized Fluid Conduits, because Ender Fluid Conduits can transmit as fast as you push into them (their built-in extract rate is bad on its own though, yeah).

GT pipes with acceptable throughput are vastly more expensive per block and you also have to worry about whether pipes will explode from heat or melt from acids or lose gas. Huge steel pipes are at best twelve ingots a piece for a mere 1800L/t, which would be enough for a lot of things but they will still bottleneck Steam in particular.

You can make EIO conduits just fine in LV (though a bit of a hassle for the plastic).

tracer4b:
I want to redo this quest completely so I'll take this over on my end. Thank you for contributing

PR #72, opened 2022-08-09
Addition of Project Red

pat2rome:
This commit integrates Project Red with the main pack. Several things have been done to make the integration seamless, and the addition leaves progression and balance fully intact. Since the wires cannot be made without red alloy ingots, there is a natural gate that places them beyond vanilla redstone (the alloy smelter).

Any duplicate and/or unused items have been removed from JEI, and GregTech silicon wafers have been substituted for the relevant recipes. The ore dictionary made it so that these substitutions were the only ones I had to do manually.

No worldgen is added, so this update causes no problems with existing worlds. The saws added by Forge Microblocks and the screwdriver added by Project Red cannot be used in place of equivalent GT tools and vice versa, so my choice to use the "screwdriver does not take damage" config does not render GT screwdrivers obsolete. All blocks that can be used for AE and GT facades have been added to microblocks.cfg so they can also be used as covers.

I aimed to make microblocks optional and as hassle-free as possible since they are for aesthetics, by using the config option for the simple framed wire recipe (vanilla sticks instead of wood strip microblocks) and by quadrupling the durability for each level of saw.


manifest.json
```diff
@@ -683,5 +683,25 @@
       "fileID": 3846179,
       "required": true
     }
+    {
+      "projectID": 229002,
+      "fileID": 2735197,
+      "required": true
+    }
```
Caedis, line 690: Looks like you are missing commas on lines 685, 690, 695, and 700

PR #85, opened 2022-08-29
Some quests requiring the wrong genesis quest

rockymoto517:
There seem to be two genesis quests, one of which being nonfunctional, that is required by all of the checkmark quests (+ 2 others) in the first questbook section. 
image
At a first look, this quest was probably not intended to be used so I just replaced all of the prerequisite quests with the normal genesis quest.
image


Caedis:
This can actually be fixed by running `/bq_admin default load` and `/bq_admin edit off` in that order

rockymoto517:
Alright thanks, didn't know that bq quirk

Issue #92, opened 2022-09-03
"Navigating Petrochem" quest does not accept "Raw Oil"


antrobot1234:
**Pack Version:**
1.4

**Description:**
The quest "Navigating Petrochem" does not accept raw oil as a satisfier. This is an issue since oil deposits give raw oil.

**Expected Behavior:**
both raw oil and oil should be accepted fluids.

**Screenshots:**
![image](https://user-images.githubusercontent.com/21179158/188276557-fb8dc1a9-4346-4a1a-ba2a-82da525cbec0.png)

okjijenAbi:
This oil can be acquired from using Oilsands Dust.

okjijenAbi:
> This oil can be acquired from using Oilsands Dust.

but otherwise, you are right.

tracer4b:
In that case, I should update this quest to accept any of the 4 oil types

IntegerLimit:
This has been fixed, please close this issue. @antrobot1234

tracer4b:
> This has been fixed, please close this issue. @antrobot1234

FYI you can put "Closes #X" in your PR and it'll auto-close the right issue

PR #102, opened 2022-09-18
More Quest Fixes

IntegerLimit:
Fixes Issue #92. See individual commits for information. (Ignore commits before Sep 18)


IntegerLimit:
Only commits that matter are commits ed27d22 and later.

tracer4b (changes requested):
CPU requiring Cleanroom is getting fixed by CEu next update (GregTechCEu/GregTech#1176) so there is no need for the CPU quest stuff

IntegerLimit:
@tracer4b PR is ready to merge.

PR #115, opened 2022-10-02
Fix Ender Chest recipe to use Steel Crate

MaxNeedsSnacks:
Since GT *chests* in CEu are instead replaced with *crates*, the recipe previously wouldn't work properly


enderger (approved):
Works in our singleplayer world.

PR #128, opened 2022-10-09
Add *nix pack mode switcher

enderger:
We basically just ported the pack mode switcher to POSIX compliant shell script (tested under OKSH). That's pretty much it, should now just be able to use the new one on OSX, GNU/Linux, and other UNIX derivatives.


tracer4b:
I don't have any Unix system suitable for testing this on, so I'm just gonna trust that it works :trollface:

enderger:
WSL/Git bash should run it fine, it’s written in POSIX compliant shell script and should work anywhere one exists.

PR #139, opened 2022-10-23
Microverse projectors fix

IntegerLimit:
Allows for muffler hatches to be placed to replace any multiverse projector casing. Removes a useless block, Multiverse Projector Vent, and fixed up some unnecessary fluid input spaces in the Advanced Microverse Projector and the Basic Microverse Projector.

The only problem so far I've encountered with this is a bad JEI/HEI preview look, with everything clumped together. However, if this needs to be fixed, it can be with a custom JEI/HEI preview, with MBT.


IntegerLimit:
Here's what the JEI/HEI preview looks like at the moment.

IntegerLimit:
@tracer4b Please review. If you feel like the JEI/HEI previews are too messy, I could add a custom one. Otherwise, the PR is ready to be merged.

tracer4b:
The JEI previews are fine

PR #167, opened 2022-12-13
Fix Issue #151 and remove conflicting Microversium EBF recipe

### LINKED ISSUE #151: No way to package Draconic Super conductor into 2x/4x/8x/16x
**Pack Version:1.4.1a with all the mods mentioned in Github.

**Description:**
There is no recipe to package Draconic Super Conductor(MAX) into 2x/4x/8x/16x, neither crafting table or the packer, so I want to know if it's the original design.

**Screenshots:**
As shown in the screenshots, there is no recipe for the Draconic Super Conductor(MAX) in a packer or a crafting table. (it's in Chinese but I think you can recognize the Draconic Super Conductor(MAX) by Voltage)

b453303f15e6cc0c46bd04577768fad

IntegerLimit:
Fixes #151

### DIFF
--- a/overrides/scripts/BlastFurnace.zs
+++ b/overrides/scripts/BlastFurnace.zs
@@ -117,8 +117,9 @@ mixer.recipeBuilder()
 	.EUt(30)
 	.buildAndRegister();
 
-////Microversium [tier 3]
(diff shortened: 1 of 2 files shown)


IntegerLimit:
I am unable to test this, due to not being able to import the zip.

IntegerLimit:
Ok, tested, with the extra .flags removed, now works as intended. Can be merged now.

PR #173, opened 2022-12-16
Fix Sterilized Filter Casing's recycling recipes

IntegerLimit:
Sterilized Filter recipes were changed, but not the recycling recipes.
Accidentally deleted the original branch, thus the new PR.


IntegerLimit:
Works as intended.
Ready to merge.

PR #186, opened 2023-01-01
Finish Processing Lines Tab & QB Rewrite P3

IntegerLimit:
Commits https://github.com/tracer4b/nomi-ceu/pull/186/commits/8291792d4c930acdcefbb8200c7bb7afacd15379 and before were merged in https://github.com/tracer4b/nomi-ceu/pull/185.

**Major Changes**
Finished Processing Lines Tab. (Not yet)

The design of all chains has been aimed to have progress going from left to right and from top to bottom.

Chains in Processing Lines Tab:
Pulsating Polymer Clay (merged in #185)
Cobbleworks (merged in #185)

The frame of Polyethylene and Polyvinyl Chloride has been added, to be finished soon.

More chains are to come, listed in TODO.

**Minor Changes: (compared to #185)**

**Both Versions:**
-Improved the description of how to turn the Prospector's Scanner to Fluid Mode, which is located in the 'Fluid Rig' quest.
-Improved the GT Electric Tools Quest, allowing any Power Unit to be accepted, as well as moving it to Early Game, as a floater
-Made GT Item Pipes quest accept any material, any size and both normal and restrictive variants, as well as setting the recommended to Tin instead of Cobalt.
-Added a note to the Extruder Quest, saying that the LV version is (mostly) useless
-Moved 'Inspector Gadget' quest to Genesis
-Reformated several quests' texts.

**NM:**
-Changed multiple mentions of Omnicoins to Nomicoins.

**HM:**
-Fixed LV Electrolyser quest, which stated that SiO2 turned into Na.
-Updated HM qb with several changes to NM qb, such as Phantom Booster quest

**Issues Fixed:**
Fixed #213 

**TODO:**

**Chains to implement:**
-Finish Polyethylene & Polyvinyl Chloride
-Add Epoxy
-Add Polytetrafluoroethylene
-Add Polybenzimidazole
-Add Polyvinyl Butyral
-Add Polyphenylene Sulfide
-Add Naquadah Processing
-Add Platline
-Add Oil Processsing
-Add Cetane-Boosted Diesel and High-Octane Gasoline
-Add Rocket Fuel

If I end up doing a port to HM, that will be in a seperate PR.

If you have any suggestions for new chains, how to change the tab, or have any general quest book issues found, feel free to comment on this PR, or make a PR into my branch.

PR #185, opened 2023-01-01
Update qb from #186, and update GTCEu, GCYM and AE2

IntegerLimit:
Also fixed all ct errors relating to GTCEu update.
This needs to be done, as I can't finish #186 before the next update.
Along with the window title change, this is pretty much all that is needed for 1.5.
Closes #189 
Closes #199 
Closes #157 by updating GTCEu
Closes #158 by updating GTCEu
Closes #166 by updating GTCEu

PR #195, opened 2023-01-11
Update GTCEu, ae2, and gcym

IntegerLimit:
Also fixed all ct errors relating to GTCEu update
This also hides all half ships in NM, although that is also done by #186


IntegerLimit:
In favor of just merging with #185

PR #207, opened 2023-01-16
Fix Diamond Sword Recipe, CT script and DME cleanup

### LINKED ISSUE #210: Recipe error in expert mode
**Pack Version:**
1.5

**Description:**
After load my save, a error info pop out in the chat
`Failed to hide recipe category: deepmoblearning.trial_keystone, caused by java.lang.IllegalArgumentException: Unknown recipe category: deepmoblearning.trial_keystone`

**Expected Behavior:**
Should no error pop out.

**Steps To Reproduce:**
Load any world in expert mode, should be able to see the error info in the chat.

**Screenshots:**

IntegerLimit:
This fixes the DME Learner original recipe not being removed.
Also, changes Diamond Sword recipe back to vanilla, as it was conflicting with the Diamond File recipe.
Fixes #210, by commenting out that line, as Trial Keystone recipes for DME aren't in JEI (yet).
Recipes for Voltage Control Units have been removed, and Voltage Control Units have been removed from jei.
Also, removes recipes made via matter in HM, by adding a config override for DME.


IntegerLimit:
There are some problems with CT scripts on HM, I'll fix that first.

PR #230, opened 2023-01-28
Add Wrench Script (from dev)

IntegerLimit:
Script is from dev.
This fixes the problem of some of Extra Utilities 2's creative items being unbreakable.

Changes from the dev script:
Thermal Creative Items were removed from list of wrenchable blocks as they are no longer obtainable, and they were known to cause problems when responding to GT Wrenches in https://github.com/Nomifactory/Nomifactory/pull/598.
Draconic Evolution Creative Energy Source was added to the list of wrenchable blocks.
GTCEu wrenches were added to the list of wrenches.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -211,7 +211,7 @@
     },
     {
       "projectID": 239197,
-      "fileID": 3587830,
+      "fileID": 4353149,
       "required": true
     },
     {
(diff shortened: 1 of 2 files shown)


IntegerLimit:
Should be good.

Issue #277, opened 2023-03-08
Compatibility script for Fluid Drawers


supasem:
I always found it frustrating that nomifactory had no way of easily managing fluids without copious amounts of super tanks and fluid storage busses

Over the last month or so, I have been playing with the [fluid drawers mod](https://www.curseforge.com/minecraft/mc-mods/fluid-drawers)
it has greatly improved organization for processing chains and has made fluid storage and interfacing with AE2 a breeze
This mod fits seamlessly with the already existing [storage drawers mod](https://www.curseforge.com/minecraft/mc-mods/storage-drawers), and drawer networks can operate with both types of drawers on the same network. (It even contains framed variants)

Simply add the [fluid drawers mod](https://www.curseforge.com/minecraft/mc-mods/fluid-drawers), and tweak some base stats to balance them with other fluid storage solutions that already exist. Tweak the recipes if needed.

![2023-01-22_22 28 33](https://user-images.githubusercontent.com/122709287/223774805-6edf390b-f24b-489c-8e97-7ca4145be100.png)

![2023-03-07_01 12 24](https://user-images.githubusercontent.com/122709287/223774870-4733f2f1-ab97-4b59-bfc4-fe85c844ee48.png)

IntegerLimit:
I'll have a think of it, and discuss with Exa and tracer.

IntegerLimit:
Okay, this will end up being a compatibility script, similar to Flux Networks, Ae2fc, and Project Red: illumination.

IntegerLimit:
Um, mostly make the recipes harder, and remove any useless items. Honestly, make the PR, and we'll request any recipe or item changes as needed.

IntegerLimit:
It has been decided that if a PR is not made by someone else, this will be done in the attached PR. The reason being, if it was implemented before this PR by me, I would just have to do it again in the PR, which just creates a waste of time.

Folanlron:
This will be super seceded by https://github.com/GregTechCEu/GregTech/pull/1963

As GTCEu is working on it's own Controller for Super Tanks/Chest..

Issue #265, opened 2023-02-28
Support for OpenComputers and Computronics


TheAirBlow:
I think adding both of those mods will be a nice addition. Maybe not out of the box, but do it in the same way as Flux Networks for example.

My thinking is that OC tiers will be locked up behind specific voltages, like tier 1 is LV, tier 2 is MV, tier 3 is HV. Additionally replace OC components with GT ones. I dunno about Computronics though, but there is a tape player which is cool xd

If I'll have time, maybe I will do a PR.

IntegerLimit:
I have no plans to make such an addon script myself, however, if someone is to PR such a script, I'm sure we could accept it.

TheAirBlow:
> I have no plans to make such an addon script myself, however, if someone is to PR such a script, I'm sure we could accept it.

Should be all tiers above LV locked behind the respective assembler tier? Is using circuit assemblers for CPUs fine? (common sense but oh well) Additionally, should I lock T3 behind the cleanroom?

TheAirBlow:
Also btw, are there CraftTweaker docs for the GTCEu?

IntegerLimit:
> > I have no plans to make such an addon script myself, however, if someone is to PR such a script, I'm sure we could accept it.
> 
> Should be all tiers above LV locked behind the respective assembler tier? Is using circuit assemblers for CPUs fine? (common sense but oh well) Additionally, should I lock T3 behind the cleanroom?

I have no idea, with no experience on Open Computers. Submit the PR, and I'll have a look if I want any changes.

IntegerLimit:
> Also btw, are there CraftTweaker docs for the GTCEu?

Honestly, just look at the CT recipe changes in the pack. But https://github.com/GregTechCEu/GregTech/wiki is also available.

TheAirBlow:
> > > I have no plans to make such an addon script myself, however, if someone is to PR such a script, I'm sure we could accept it.
> > 
> > 
> > Should be all tiers above LV locked behind the respective assembler tier? Is using circuit assemblers for CPUs fine? (common sense but oh well) Additionally, should I lock T3 behind the cleanroom?
> 
> I have no idea, with no experience on Open Computers. Submit the PR, and I'll have a look if I want any changes.

Think of CPUs like GregTech circuits. They're constructed in almost the same way - with transistors and other kinds of stuff. Components are also tiered, excluding the basics lime Transistors.

IntegerLimit:
[...]

Issue #134, opened 2022-10-15
Ae Additions Addon Script


adamcirillo:
are you able to make an addon script for AE Additions?

https://www.curseforge.com/minecraft/mc-mods/ae-additions-extra-cells-2-fork

tracer4b:
No plans to, if someone would contribute such an addon script, I would have no issue accepting it.

IntegerLimit:
It has been decided that if a PR is not made by someone else, this will be done in the attached PR. The reason being, if it was implemented before this PR by me, I would just have to do it again in the PR, which just creates a waste of time.

If you do make a PR for this, I would recommend you PR a Groovy Script (Not CT) to the `groovy` branch.

Folanlron:
Stale

IntegerLimit:
Part of plan of 1.8.

HandsomeAsianMan:
I think nae2 might have made this irrelevant

IntegerLimit:
AE Additions does have other features, will check if should make such a script later.

D-Alessian:
Just looked into it, it adds a poor man's version of AE2FCr. All this would add to Nomi is an oredict export bus, an ME drive cover, an ME cell cover, 65m drives, a blast resistant ME drive as well as "ME fluid autofiller" and "Fluid vibration chamber" which I couldnt figure out a use for. Only thing I could imagine would be to make a script similar to AE2FCr that wouldn't load if AE2FCr is loaded but I think it's safe to say an addon script would be pretty useless for this

IntegerLimit:
Closing. Use AE2 FCr + NAE2's extra cells instead. Gas support is not needed as we do not have Mekanism.

PR #291, opened 2023-03-17
CraftTweaker -> GroovyScript

IntegerLimit:
# Original CraftTweaker -> GroovyScript PR

This PR is going to be *half* scrapped. https://github.com/Nomi-CEu/Nomi-CEu/pull/579 contains some of this, but the other scripts will be transitioned soon, either after a large force-push to this PR, or a new PR.

The *complex* stuff made here has already been transitioned to Nomi Labs.
This branch is being kept for the time being, hopefully for use in the GroovyScript Community, and with the next PR.

See Below for the Original PR Description. Some parts of the roadmap may be followed for the new PR.

~~~
## Link to GroovyScript:
[Curseforge](https://www.curseforge.com/minecraft/mc-mods/groovyscript)
[GitHub](https://github.com/CleanroomMC/GroovyScript)

## Disclaimer:
This PR is nowhere near complete. Many things still may change. 

Especially: MBT being replaced with GS is being tested, hopefully, worlds should be able to be transferred easily, using the same id as the original controllers of the old scripts. Otherwise, MBT and MBT related scripts will stay.

Another small thing is that Pack Mode may be removed, depending on how the final implementation of mode switching ends up being.

## TODO, in the current planned order:
- [x] Add Groovy to Modpack, remove CT, ContentTweaker, and MBT
- [x] Remove all CT scripts
- [x] Add Custom Items without complex logic
- [x] Add Custom Blocks that are simple
- [x] Add Excitation Coil and Dust Block
- [x] Make Excitation Coil Rotatable (Like Pistons)
- [x] Add Smores
- [x] Add Hand Framing Tool
- [ ] Hand Framing Recipes
- [x] Add custom GT materials (From both modes)
- [ ] Add custom GT Multis (From both modes)
- [ ] Common scripts/functions, such as `common.zs`
- [ ] OreDicts, and items hidden from JEI/HEI, that are common to both modes
- [ ] Recipes, common to both modes
- [ ] Switching between NM and HM
- [ ] Stuff, not including Custom Multis, that are specific to one mode. 
- [ ] Final Scripts, that aren't addon scripts, not moved over/implemented
- [ ] Addon Scripts
- [ ] Cleanup
- [ ] Add Fusion Recipes & Ender IO Splice Recipes directly, rather than through config
- [ ] Fluids

## New Features to add:
- [x] Ultimate Generator and Power Source to be placeable, with rarity
- [x] GCP Greenhouse (solve apple problem, suggested by @GGabrielDev)
- [ ] Maybe new multis, such as MK IV (for HM mainly)
- [ ] #134 
- [x] #277
- [ ] #265 
- [ ] Access https://github.com/Nomi-CEu/nomi-ceu/issues/345
- [ ] Have the nbt edited terminal be given to players on their first log in, instead of via qb (use groovy native function, doesn't matter if they get given one again)
- [ ] Contributing Docs
- [ ] New modlist.html, which goes into created client zip
- [ ] Maybe have Ultimate Generator and Power Source actually produce/store power
- [ ] Add command to show server information, change island cake tooltip to tell people to run command

## Plan for custom fluids:
If groovy implements an api for fluids, then we'll use that
Otherwise, we make the fluids gt materials
~~~


IntegerLimit:
Scrapped, to be replaced by a up-to-date PR.

PR #335, opened 2023-04-17
Update buildscripts

Exaxxion:
Merged changes from https://github.com/Nomifactory/Nomifactory/tree/dev/buildtools made after February 2022 (when I helped @tracer4b get the build scripts at the time set up).

CurseForge subsequently did its API changes so the old scripts don't work anymore.

You will need to create a Secret called `CFCORE_API_TOKEN` which contains the CurseForge API authentication token you wish to use (for https://api.curseforge.com). When running build scripts locally, you need to have an environment variable set with the same name and value.

This token allows the script to download mod jars required for building the server zip. Without this token, at the very least mods that disallow third-party downloads would return blank download links and that build target would fail. It might be needed to use the API at all, but @notmywing is our CICD author and resident expert, so I would defer to him on details.

### DIFF
--- a/.github/workflows/deploycfrc.yml
+++ b/.github/workflows/deploycfrc.yml
@@ -1,0 +1,83 @@
+name: Deploy to CurseForge (RC)
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to checkout'
+        required: true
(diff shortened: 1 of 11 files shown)


IntegerLimit:
Looks good! Thanks!

PR #372, opened 2023-05-31
Added Island Cake configs

Caedis:
Disabled by default
Target dim = void world (119)


overrides/config/DimensionalEdibles.cfg
```diff
@@ -314,6 +317,42 @@ dimensionaledibles {
...
+
+        islandcake {
+            # Allow players to create personal islands when on a FTB Team
+            B:allowPersonalIslands=true
+
+            # Set to true to make the Island Cake consume fuel.
+            B:consumesFuel=true
+
+            # Set the fuel used by Island Cake (Don't change this unless you know what you're doing).
+            S:fuel=minecraft:cobblestone
```
IntegerLimit, line 335: Can this be changed to pulsating dust?

overrides/config/DimensionalEdibles.cfg
```diff
@@ -20,6 +20,9 @@ dimensionaledibles {
 
         # Set to true to enable Ender Apple.
         B:enderApple=false
+        
+        # Set to true to enable Island Cakes.
+        B:islandCake=false
```
IntegerLimit, line 25: Should this be true?

IntegerLimit (changes requested):
Can you add this information to the quest book?

overrides/config/DimensionalEdibles.cfg
Caedis: Yep, just change it to the internal name

Caedis:
The best way to handle is for the server admin to enable them, use the default void cake to go to 0,0 in the void dim, then place down some infinite use island cakes, including some personal island cakes.

Caedis:
Personal island cakes are made by right clicking a normal island cake with the `personalLockingItem` item

IntegerLimit:
It would be nice if you could change the fuel quickly, and add a README or qb so that the server admin can easily know about this. Otherwise, its great!

IntegerLimit (approved):

PR #374, opened 2023-06-04
Extended crafting recipe tweaks

BlakeTech:
Changes a few recipes in Extended Crafting to fix/allow NBT-insignificant similar items to be used.
 Specifically, Watering Cans Range setting is ignored, and the log type required for the Skullfire Sword is no longer specific to Oak.


IntegerLimit:
All okay, I'm assuming this superseeds the previous PR, great.

overrides/scripts/extendedcrafting.zs
```diff
@@ -720,10 +720,10 @@ makeExtremeRecipe9(<thermalcultivation:watering_can:32000>,
      "    CCC  ",
      "         "],
     { A : <thermalcultivation:watering_can>.withTag({Water:0, Mode:0}, false),
-      B : <thermalcultivation:watering_can:1>.withTag({Water:0, Mode:0}, false),
-      C : <thermalcultivation:watering_can:2>.withTag({Water:0, Mode:0}, false),
-      D : <thermalcultivation:watering_can:3>.withTag({Water:0, Mode:0}, false),
-      E : <thermalcultivation:watering_can:4>.withTag({Water:0, Mode:0}, false),
+      B : <thermalcultivation:watering_can:1>.withTag({Water:0, Mode:1}, false),
```
IntegerLimit, line 723: Why does changing this fix the problem?

BlakeTech:
A bit more clear explanation in that image, perhaps.

IntegerLimit:
It would be better to use the CT function to have any NBT. I'll do that shortly, and see if it is possible, and if it works well. If not, this seems like a good solution.

BlakeTech:
Sure. You can ping me on Discord if it works, and I'll close this PR instead. The changes are small enough you can just copy/paste them in.

IntegerLimit:
I'll probably just make commits to your branch, and then merge this PR.

IntegerLimit:
Also, why are the commits made by someone else?

BlakeTech:
My alt account. I have two, and sometimes, IDEA just doesn't remember which one I chose for it to push it as.

IntegerLimit:
Ah, okay, that makes sense.

IntegerLimit:
Sorry for the delay, I've been a bit busy lately... I'll look at this today or tomorrow.

IntegerLimit (approved):

PR #378, opened 2023-06-06
Update NNP to 1.1.0 and add translation keys

NotMyWing:
Updates NNP to 1.1.0 to support notifying nearby players why Nether portals are disabled.

image


IntegerLimit:
Thanks! That looks great!

PR #387, opened 2023-06-29
Better Chinese translation

mczph:


IntegerLimit:
I don't have the resources to check this, but looking at your many Chinese localization commits in so many repos, I'll trust you and merge this.

PR #400, opened 2023-07-22
Update DefaultQuests.json for Expert / Hardmode

BeardBois:
In expert/hardmode it seems that potin pipes transfer rates were changed compared to normal mode. So I changed the number for the correct number. 

Was 414 mB/t changed to 192 L/t

I think that I did this correctly.

### DIFF
--- a/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
+++ b/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
@@ -3059,7 +3059,7 @@
       "properties:10": {
         "betterquesting:10": {
           "autoclaim:1": 0,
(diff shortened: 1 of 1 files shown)


IntegerLimit:
Can you check if this inconsistency happens in NM too?

IntegerLimit:
> In expert/hardmode it seems that potin pipes transfer rates were changed compared to normal mode. So I changed the number for the correct number.
> 
> Was 414 mB/t changed to 192 L/t
> 
> I think that I did this correctly.

It seems that you changed the whole questbook? What did you do?

IntegerLimit:
Superseeded by https://github.com/Nomi-CEu/nomi-ceu/commit/a74005c830fe220c7307ef4bc08432af6aa5379c.

This PR is no longer needed. However, thank you for bringing this issue to my attention.

PR #401, opened 2023-07-23
Add NAE2 and Pattern Mult-Tool quest, update MixinBooter to 8.3

NotMyWing:


IntegerLimit:
Thank you. Also:
- You are the only PR from another person that has actually properly changed all 4 qbs
- You left the manifest.json sorted!!!

PR #414, opened 2023-08-27
Enable shufflemode on main menu

ThePansmith:
Single line PRs :trollface:, anyways, this simply just enables shuffle mode on the main menu. As is the images play in a fixed order.


IntegerLimit (approved):
Works extremely well. :+1:

PR #431, opened 2023-09-16
Allow for QB Translations of Normal and Expert Quest Books

IntegerLimit:
Also moves all nodejs stuff into `tools` package.

Also adds a basic `CONTRIBUTING.md` file.


IntegerLimit:
TODO: 
- Make it only run on master repo.
- Remove `qb-lang-revamp` branch from the events
- Allow workflow to run on all branches starting with `test_qb*`?

PR #444, opened 2023-09-28
Clear up one-way power conversion in `Cables` Quest

sparr:


IntegerLimit (changes requested):
As said in Discord, please add these changes to the Expert QB as well.

sparr:
Added to expert QB. Also fixed an apostrophe encoding mistake from #437 per discussion.

IntegerLimit (approved):

PR #445, opened 2023-09-30
Move Genesis quests, eliminate overlapping lines

sparr:
The vertical prerequisite lines between Electronics Subcomponents, Primitive Steelworks, and Your First Circuit overlap in an ambiguous way. This change moves two of them to avoid that.
image

This problem does not exist in the expert quest book.


IntegerLimit (commented):
Can you include a screenshot of what it looks like now? Otherwise, looks good.

sparr:
Current qb layout with the overlapping lines
image

IntegerLimit:
> Current qb layout with the overlapping lines image

This is the current layout, correct?

sparr:
Yes

IntegerLimit:
Sorry, but I think the new version looks too messy. The old version is fine, as the quests, while not dependent on each other, are still related. If you have further suggestions/ideas, comment on this PR, or make a new one.

PR #458, opened 2023-10-19
Fix Mac/Linux Pack Mode Switcher Erroring

uku3lig:
if `server.properties` is not present in the minecraft directory, it will try to delete it which will fail and exit the script due to `set -e`


IntegerLimit (approved):
Approved, but waiting for the person who wrote those lines to quickly look over it.

0xC4DE:
Small nitpick -- the rm was just to insure it's gone, you can eat error codes with `|| true`, Alternatively, removing `set -e` could have been a better option. 

This works, too. Thought I started with `cp`, hence the `rm` after the copy, but it got changed to `mv` at some point.

IntegerLimit:
@uku3lig are you going to implement the suggestions above?

uku3lig:
my bad, i misunderstood what 0xC4DE meant. i added what was suggested

IntegerLimit:
> my bad, i misunderstood what 0xC4DE meant. i added what was suggested

Sorry if I'm misunderstanding, but:
Your previous commit was an error
Where is the fix?

uku3lig:
1091b039090124f3ee822fc9d97ae0431f9c8759 (latest commit in branch) is the correct commit and has the modifications suggested

i should've sent my comment before pushing it, sorry

PR #478, opened 2023-10-24
Allow Changing of Sound Output Device

### LINKED ISSUE #477: add Sound Device Options
### Nomi CEu Version

1.7-alpha-1

### Related Problem

minecraft audio can just break due to changing of sound devices

### Your Solution

addding [this mod](https://www.curseforge.com/minecraft/mc-mods/more-sound-config) allows for me to change the audio device without having to restart the game or reload textures (which can cause problems)

we add a tiny mod, and GregTech Community Pack should have it included so you can try it out there ig

idk pesting the random labs to add it to one of their mods instead

### Additional Information

Whenever my pc goes to sleep, my sound device is lost for the duration of the sleep, so when i wake it up i lose all sound in minecraft because is such an amazingly programmed game

htmlcsjs:
closes #477

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -759,6 +759,14 @@
       "projectID": 896591,
       "fileID": 4728250,
       "required": true
+    },
+    {
+      "projectID": 306555,
+      "fileID": 2707092,
+      "required": true,
+      "sides": [
+        "client"
+      ]
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


IntegerLimit (changes requested):
Manifest entries are sorted by projectID. Please place this one in a proper spot, instead of just at the end of the file.

IntegerLimit (approved):

PR #548, opened 2023-12-02
Chinese Translations for AE2, DME and Normal QB

### LINKED ISSUE #532: Language Pack for Chinese Players
### Nomi CEu Version

1.7-alpha-1

### Related Problem

_No response_

### Your Solution

For personal use, I [fork](https://github.com/zero6six/nomi-ceu-zh_cn) a repo from IAmNotGEM (#164) and combine it with official langguage pack to a resourcepack which is suitable for chinese players.It contains updated Chinese language file for GTCEu (provided by a [paratranz project](https://paratranz.cn/projects/7760), because official version is outdated) and questbook (not contain expert mode and some chapters, bilingual quest description).

Aiming to share it with more Chinese players, I open this issue. And nomi-ceu is popular and renowned among Chinese hardcore players but lack stable localization for latest version.

The main translated questbook text before version 1.6 is from IAmNotGEM with high quality.After that, I translate new quest and a few changed old ones by machine translation and polish.But I swear that the translated text can be understand by players and be true to the original.I also provide bilingual description for questbook. And that is all.

Releases from my repo function as normal resourcepack.Just need to load it in the game.

For Chinese Players, I have some optional advices:
1. load this languagepack with a [universal languagepack](https://github.com/CFPAOrg/Minecraft-Mod-Language-Package)(needing to remove gregtech file, for it provide wrong files(because GTCE,GTCEu, etc. use the same modid))
2. if necessary, remove serializationisbad mod or use proxy because it generate crash in downloading files, which is blocked by GFW.

### Additional Information

The main object of this issue is Chinese players.
**If someone was unhappy about that or I disturb someone, I apologize for that.**

zero6six:
https://github.com/Nomi-CEu/Nomi-CEu/issues/532#issuecomment-1836315458

Fixes #532 

Since I didn't receive a reply, the original text of quest is not attached to the translation.


IntegerLimit:
> [#532 (comment)](https://github.com/Nomi-CEu/Nomi-CEu/issues/532#issuecomment-1836315458)
> 
> Since I didn't receive a reply, the original text of quest is not attached to the translation.

What do you mean by bilingual?

Note that if the translation key is not present in the file, it defaults to the en_us one, even if the language is set to Chinese.

If you mean including en_us.lang, yes, that is not needed.

Apart from this, approved.

IntegerLimit (approved):

zero6six:
Bilingual means that I copied the original task descriptions from en_us and inserted them at the end of the description in zh_cn. I do this in bulk with a script.

The effect looks like this.





I'm not sure whether I should do it, because it has both advantages and disadvantages.

IntegerLimit:
This does seem like a good idea in some ways, but I think I'll merge this PR as-is.

If we were to implement this, I would rather have the English part auto-update or something, and we would also have to do this for other translations.

Can you make an issue so we can discuss this issue later? (post 1.7)

PR #584, opened 2023-12-21
Create a modlist markdown file to find mod better

Darknight123MC:
Draft because some mods can't write page URL by metadata.
I exported it by `PrismLauncher` and `1.7-alpha-2a`


IntegerLimit:
This is already planned. Currently, in any instance, you can double click on `modlist.html` to see that.

We plan to make a github pages site with changelogs and modlist, although that is in progress.

The reason we need to set it up is so that it can be auto-updated.

Thanks for the PR!

PR #623, opened 2024-02-18
Replace §f with §r in Secondary NM QB & HM QB

IntegerLimit:
Essentially, I completely messed up https://github.com/Nomi-CEu/Nomi-CEu/pull/615.

This PR:
- Updates the other Normal Mode Quest Book File
- Replaces §f with §r in both HM Quest Book Files


IntegerLimit:
I really need to write a GHA Check for this...

Issue #645, opened 2024-03-07
Crash when taking liquid out of an EnderIO portable tank with a bucket


lynxx131:
### Nomi CEu Version

Nightly

### Launcher

Prism Launcher

### Configurations or Mods Changed

- Added TConstruct-1.12.2-2.13.0.183
- Set LoliASM onDemandAnimatedTextures to false

### Environment

Client and server both running devgt-2.8-d588519-client

### Related Problem

I'm using this template as GitHub says the Bug Report template YAML contains a syntax error

### Your Solution

N/A

### How Will Your Solution Work

N/A

### Additional Information

When using a bucket to take fluid out of an EnderIO fluid tank the game crashes with the attached crash report [crash-2024-03-07_17.55.36-client.txt](https://github.com/Nomi-CEu/Nomi-CEu/files/14520781/crash-2024-03-07_17.55.36-client.txt).

Breaking the tank in creative also causes a crash, reverting to Alfheim 1.1.1 seems to fix this issue.

lynxx131:
I've raised https://github.com/Red-Studio-Ragnarok/Alfheim/issues/42 so hopefully should get sorted.

lynxx131:
@IntegerLimit Just wanted to add to this I've been running Alfheim 1.3 which fixes the issue I logged for a bit now and I've not had any issues so not sure if you want to look at bumping to 1.3 in the next release.

PR #649, opened 2024-03-09
Downgrade Alfheim from 1.2 to 1.1.1

IntegerLimit:
Fixes https://github.com/Nomi-CEu/Nomi-CEu/issues/645.

Will Upgrade Again when it is stable.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -773,7 +773,7 @@
     },
     {
       "projectID": 910715,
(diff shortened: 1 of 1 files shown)


IntegerLimit:
It was also randomly crashing launches and crashing `/gs reload`, which is the cause for the downgrade.

PR #655, opened 2024-03-13
Adding a Blackhole to the space dimension

almajd3713:
I kind of wished there was a blackhole background for my space base (well, future), so I just added that for the planetDefs file. 
You can change the settings for it to make it more distant and consume more fuel for example, or just to change its name.

### DIFF
--- a/overrides/config/advRocketry/planetDefs.xml
+++ b/overrides/config/advRocketry/planetDefs.xml
@@ -1,5 +1,5 @@
 <galaxy>
-    <star name="Sol" temp="100" x="0" y="0" numPlanets="0" numGasGiants="0">
+	<star name="Sol" temp="100" x="0" y="0" numPlanets="0" numGasGiants="0">
 		<planet name="Earth" DIMID="0" dimMapping="">
 			<isKnown>true</isKnown>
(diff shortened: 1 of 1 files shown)


overrides/config/advRocketry/planetDefs.xml
```diff
@@ -112,4 +112,5 @@
 			<rotationalPeriod>21100</rotationalPeriod>
         </planet>
 		</star>
-</galaxy>
\ No newline at end of file
+    <star name="Eye Of Eternity" numGasGiants="0" numPlanets="0" size="2.0" temp="800" x="300" y="120" blackHole="true" > </star>
```
IntegerLimit, line 115: This should be a tab indentation, to fit with the rest of the document.

almajd3713: My bad, my editor uses spaces instead of tabs which I will change now. Thanks for the heads up!

almajd3713:
Okay uh as it turned out, Sol does indeed have a blackhole background for some reason. Although in my initial testing I am 120% sure there wasn't anything but a void there. Closing this pr. Thanks for your time.

IntegerLimit:
Does it? Where?

PR #658, opened 2024-03-14
Tweak Universal Crystallizer Recieps

ghzdude:
The recipes Eu/t been tweaked to be below 1 amp of UHV (2,097,152 EU/t). Eu/t and durations have also been balanced to make the EU total match the RF totals of DE fusion and combination crafting recipes,


IntegerLimit:
Thanks!

PR #662, opened 2024-03-17
Slightly Increase EB Reach

ProTriforcer:
This should make max EB shape sizes powers of 2.
It shouldn't upset balance much (not as much as array mining anyway), but will make the mod minecraftier and building chunk-aligned platforms less annoying.

The numbers look kinda random and i can't figure out what the formula is, but they should work.
Here are max floor sizes for each level before:
image
image
image
image

And after: 
image
image
image
image

### DIFF
--- a/overrides/config/EffortlessBuilding.cfg
+++ b/overrides/config/EffortlessBuilding.cfg
@@ -8,21 +8,21 @@ reach {
 
     # Maximum reach in creative
     # Keep in mind that chunks need to be loaded to be able to place blocks inside.
-    I:maxReachCreative=200
+    I:maxReachCreative=211
 
     # Maximum reach in survival without upgrades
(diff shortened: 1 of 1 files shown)


IntegerLimit:
The formula compares:
The Horizontal Distance of the two corners (Calling that `x`)
The Reach (Calling that `y`)

It takes the largest `x` that can fit in `y*4`.

So for reach 20, that's a maximum of 80, of which the closest square distance to that is 6 (6^2 * 6^2 = 72).

IntegerLimit:
* Which is really stupid in my opinion.

IntegerLimit (approved):

IntegerLimit:
Nevermind.

Formula:
Max blocks per axis: `reach * 0.3`
Max blocks placed at once: `reach * 1.6`

Issue #675, opened 2024-03-24
Allow Clay production using Distilled Water


a-a-GiTHuB-a-a:
### Nomi CEu Version

1.6.1b

### Launcher

CurseForge (doesn't matter)

### Configurations or Mods Changed

added AE2FC (doesn't matter)

### Environment

Singleplayer (doesn't matter)

### Related Problem

I can't use the Distilled Water I get from my large steam turbine to make clay from dust, even though the distilled-ness shouldn't affect how the clay is made.

### Your Solution

Allow using distilled water to make clay!

### How Will Your Solution Work

add a recipe to make clay with dust and distilled water, or use some kind of water oredict

### Alternatives You've Considered

use regular water from elsewhere and trash the stuff from the big turbine

IntegerLimit:
Are you sure this is still an issue in 1.6.1b?

I swear it was fixed ages ago (GT 2.5.1) https://github.com/GregTechCEu/GregTech/pull/1483...

IntegerLimit:
Closed as not reproducible.

<img width="456" alt="image" src="https://github.com/Nomi-CEu/Nomi-CEu/assets/103940576/d5c8e3b4-61f4-414a-b0f9-02e10963e43a">

rogerwim:
> Closed as not reproducible. (Both NM and HM, 1.7-alpha-3)
> 
> 
> 
> <img width="456" alt="image" src="https://github.com/Nomi-CEu/Nomi-CEu/assets/103940576/d5c8e3b4-61f4-414a-b0f9-02e10963e43a">
> 
> 

This isn't what was meant, I am fairly sure the chem reactor recipe for dust BLOCK into clay BLOCK is the issue

a-a-GiTHuB-a-a:
I meant the dust _block_ in cobbleworks, not clay dust.

IntegerLimit:
Yep, fixed in the above PR.

PR #685, opened 2024-03-26
Add Distilled Water Variant of Dust -> Clay

IntegerLimit:
Fixes https://github.com/Nomi-CEu/Nomi-CEu/issues/675


tracer4b (approved):
I think there's a certain reduction % that distilled water recipes are supposed to have over water... Too bad!

IntegerLimit:
> I think there's a certain reduction % that distilled water recipes are supposed to have over water... Too bad!

There is normally, but this is what they did for the clay ball distilled water recipe as well.

PR #694, opened 2024-03-27
Fix dust-to-dye recipe conflicts

ProTriforcer:
Crafting dusts into dyes no longer conflicts with uncrafting dusts into piles, but requires a crafting table


IntegerLimit (approved):

IntegerLimit:
I have a slightly better idea. Returning to this later.

IntegerLimit:
Superseeded by https://github.com/Nomi-CEu/Nomi-CEu/pull/696.

Issue #472, opened 2023-10-23
1.7 TODO


IntegerLimit:
### Nomi CEu Version

1.6.1a

### Related Problem

I'm always frustrated when I forget what to do.

### Your Solution

I could make a TODO! Yes me. I will make it happen, no need to like to have it happen.

### How Will Your Solution Work

Make an issue, thats a TODO!

### Alternatives You've Considered

Make a PR, thats a TODO!

### Additional Information

TODO:
- [x] Increase the amount of Neutronium from T9 Micro Miner
- [x] Decrease the amount of Iridium required for building Research setups (NM only)
- [x] Add JEI Descriptions for Various AE2 Items (Sticky Card...)
- [x] Add Betterer P2P
- [x] Update AE2 UEL
- [x] Move from AE2 Stuff to AE2 Stuff Unofficial
- [x] Update NAE2
- [ ] Update Alfheim when it is stable (1.2 is not stable)
- [x] Update Mixinbooter
- [x] Update Universal Tweaks
- [x] Update Vintage Fix
- [x] Update Other Mods
- [x] Update LoliASM

v3ect0rgames:
It works if you 
Ae2 unofficial extended life to 0.56.6
Ae2stuff unofficial to 1.12.2-0.9
and delete the remove recipes for the old encoder on lines 163, 166 in AdvRocketry.zs

v3ect0rgames:
Universal tweaks seems to work on 1.12.2-1.11.0

v3ect0rgames:
Alfheim 1.3 seems to be stable

v3ect0rgames:
Thats 3 things on the to-do list

IntegerLimit:
> Alfheim 1.3 seems to be stable

It consumes a lot of ram, and thus we can't use it.

IntegerLimit:
Also, everything is already confirmed to work, its just what needs to be done. However, thank you for your testing.

v3ect0rgames:
ok thanks,  what alfheim version will be used then

IntegerLimit:
> ok thanks, what alfheim version will be used then

The current version, it uses an appropriate amount of RAM.

IntegerLimit:
This issue is now being closed, as Alfheim update uses too much memory for Nomi-CEu, and results in no improvements.

PR #712, opened 2024-04-03
Mod Updates As Specified In #472

HandsomeAsianMan:
## Whats done   
1. Betterer P2P  
2. AE2 UEL update  
3. Changed ae2stuff to ae2stuff unofficial   
    1. Encoder was removed which affected the recipe for packaged auto encoder
    2. For now i have just replaced it with hv hull so the script works
4. Updated Nae2   
    1. Haven't added recipes for the anything yet  
5. Updated Alfheim  
6. Updated universal tweaks  
7. Updated vintage fix  
8. Updated loliasm  

## Needs done
1. AE2 stuff encoder replacement
2. Recipes for extra stuff added by nae2


IntegerLimit:
Please update Had Enough Items as well, otherwise Alfheim kills memory.

Furthermore, can you please undo the update to Universal Tweaks?
There are some config options that I would like to try out.

For AE2 Stuff Encoder, there might need to be more done.

There also needs to be a JEI description page for AE2 Sticky Cards, and Quests for NAE2 (Although that can be in a seperate PR)

HandsomeAsianMan:
[Updated mods from nomi 1.7.4.txt](https://github.com/Nomi-CEu/Nomi-CEu/files/14872194/Updated.mods.from.nomi.1.7.4.txt)
found some other mods to consider updating aswell

HandsomeAsianMan:
FluidCells
nae2Cells

HandsomeAsianMan:
For the sticky card. Idk what to actually do by making a jei page for it but here is the description.  

Sticky Card is used to define exclusive destinations for items/fluids storage making it so all items inputted into the system will be stored there regardless of priority. If the designated storage is full, the designated item will NOT go anywhere else even if space is available elsewhere.

IntegerLimit:
I will make the JEI desc, leave that to me in a future PR (which will include NAE2 quests)

HandsomeAsianMan:
Aight cool this is basically done than if ur happy with the cell recipes n stuff

IntegerLimit:
I am also concerned about the removal of the ae2 stuff pattern encoder. What will happen to existing items in pattern encoders? Can Pattern Encoders store items? What would be the closest full block that is equal to it?

HandsomeAsianMan:
The only item that u c store in are patterns I think. Up to a stack, the recipe slots are filled with fake items. So maybe some people will lose those if they have an encoder placed around. They also count as an ae2 network part and some people might have one placed around just to connect up their ae2 system instead of a cable or other block. That's about it. So the closest full block is probably just a full block interface?

IntegerLimit:
As requested, I will perform some formatting changes on this PR.

IntegerLimit (approved):

IntegerLimit:
Balancing Changes:
256k cells are now iv, and are assembler recipes
image

Changes in EUt for 4096k and 16384k
image
image

tracer4b (approved):

PR #720, opened 2024-04-10
Various QB changes

SparkedTheorem:
-changed Research Station quest to require Trinium Coils quest due to presence of Uranium Rhodium Dinaquide Wire which requires Trinium Coils to smelt
-changed T9MM quest description to state that it brings in a large amount of Neutronium rather than a decent amount
-changed Large Circ Assembler quest description to note that it cannot accept more than one energy hatch
-changed IV Assembling Machine quest description to note that the GCYM Assembler can't take more than one energy hatch
-changed PBI quest description to sound less scary and note that there are some steps you can skip
-changed all wafer quests to have same task formatting (two retrieval tasks with taking either uncut or cut wafer, uses or logic)
-changed World Accelerator quest to require Research Station since World Accelerator is gated behind it
-added information to Indium quest regarding the circuit 4 recipe since it skips a step
-replaced Enderium Ingot quest icon with nomi labs version of Enderium
-Marked Breaking Some Eggs quest as main quest since its definitely not optional
-changed Optical Fiber Cables quest to also require Boron quest to be completed since you need Borosilicate Glass
-added line about Red Granite in Uranium Fuel Infrastructure quest
-updated Active Transformer quest to require Trinium quest to be completed first since you need a Superconducting Coil block

Got no clue why so much stuff was deleted from the override json, I copy-pasted from my development instance of the Nightly build since that's what the wiki says.


overrides/config-overrides/normal/betterquesting/DefaultQuests.json
IntegerLimit, line 1: Please undo this change for now, this will be performed at the end of the qb development.

Just leave the config override version as it was before.

Otherwise, looks good!

IntegerLimit (changes requested):

overrides/config-overrides/normal/betterquesting/DefaultQuests.json
IntegerLimit: You did the right thing, copying to the config-override version, but this is not required for this specific quest book development branch.

SparkedTheorem: Reverted.

IntegerLimit (approved):

Issue #741, opened 2024-05-14
Add PackagedDraconic


SparkedTheorem:
### Nomi CEu Version

1.7-alpha-4

### Launcher

Prism Launcher

### Configurations or Mods Changed

Updated LoliAsm to 5.20, Updated AE2 UEL to v0.56.0, Updated NAE to 1.6.1
Added AE2 Fluid Crafting Rework, Added Flux Networks, Added Lazy AE2, added PackagedDraconic (lol)
Modified EnderIO to make Telepad not take fluid to use, modified Rod of Return to be cheaper and faster to use as well as make it not use fluid

### Environment

Singleplayer

### Related Problem

Automating Draconic Evolution Fusion in the endgame sucks, between Actually Additions Lasers not cooperating and fusion injectors randomly messing up I had to perform maintenance on my fusion setups far more than I desired.

### Your Solution

PackagedDraconic, an addon mod to PackagedAuto and Draconic Evolution, provides a significantly easier way to automate Draconic Evolution Fusion. It should get added into the modpack.

### How Will Your Solution Work

The Packaged Fusion Core and Marked Fusion Injectors should get added with unmodified recipes most likely, as there isn't much of a good reason or viable way (imo) to gate convenience of automation when you're expected to basically do it immediately.

DraconicFusion would replace the AA laser automation suggested in the QB.

v3ect0rgames:
Version 1.12.2-1.0.0.13 of PackagedDraconic seems to work in 1.7 alpha 4 although the energy used might need balancing in the .cfg file

PR #769, opened 2024-06-17
Implement #741

v3ect0rgames:


IntegerLimit:
This needs to be added with balancing, as well as potential changes and tweaks in Nomi-Labs, and the branch here is not based off the correct branch (Nomi-CEu/main). Closing this for now.

PR #772, opened 2024-06-19
Add Lategame GT Recipes for AE2 Components 

### LINKED ISSUE #754: Gregify AE2 recipes for late game
### Nomi CEu Version

1.7-alpha-4

### Launcher

Prism

Only ones relevant are :
AE2FCr
Lazy AE2

### Environment

Forge server

### Related Problem

Currently, late game, making anything anything AE2 related is particularly painful. Lazy AE2 somewhat fixes this with circuit etchers and pulse centrifuges, but it isn't in the pack by default.

### Your Solution

Make recipes for AE2 components (processors, purified quartzes, charged quartz, fluix crystals) have crafts in GT machines.

> Purified (certus quartz, quartz, fluix) : Use the new greenhouse multi, and have the same recipe as the phytogenic isolator recipe for them. Gate it at EV and make it take 2 seconds so it takes 1t at UHV.

> Processors : Have the same recipes as the circuit etcher recipes, but in a forming press, I think it's the machines that makes the most sense. Gate it behind IV, recipe speed doesn't matter as much since it can parallel

> Fluix : Make a mixer recipe for fluix dust that uses `charged certus quartz dust, nether quartz dust, redstone` that makes 2 fluix dust. Have an autoclave recipe that make fluix crystals from fluix dust and water/distilled water. The gating on this should be around EV too I feel like, and speed again is less important since parallels.

> Charged certus quartz : This one, I'm not entirely sure what recipe to have for it, but maybe electrolyser? It should be the least gated though I think, it's the least "precious" AE2 material. Maybe HV with a very low recipe time, like 0.5s.

Adding Lazy AE2, but it still doesn't feel the same. I use Lazy AE2 on my world, still feels slow at post tank where I'm at.

Having custom machines made for AE2 recipes that are gregified would also be a better alternative imo, but it may be too much work for a limited impact.

### Additional Information

I think it's important to gate them pretty late to not make it too easy. But late game, AE2 is necessary, and you have pretty much no good options to make a ton of AE2 items that isn't painful. That pain is what I'm proposing to fix, not making AE2 less gated and easier.

_No response_

tracer4b:
Adds GT recipes for AE2 processors, Fluix, and purified crystals.

Fixes #754.

Preview of recipes:
Screenshot 2024-06-19 at 10 02 58 AM
Screenshot 2024-06-19 at 10 04 49 AM
Screenshot 2024-06-19 at 10 06 55 AM

### DIFF
--- a/overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/items.groovy
+++ b/overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/items.groovy
@@ -213,3 +213,55 @@ crafting.shapedBuilder()
 	.key('C', item('appliedenergistics2:material:56'))
 	.key('D', ore('circuitEv'))
 	.replace().register()
+
+// AE2 Processor Shortcut
+// Logic Processor
+mods.gregtech.circuit_assembler.recipeBuilder()
+    .inputs(ore('circuitEv'), metaitem('plate.nand_memory_chip') * 8, metaitem('wireFineElectrum') * 16, item('appliedenergistics2:part', 16) * 2)
+    .fluidInputs(fluid('soldering_alloy') * 288)
+    .outputs(item('appliedenergistics2:material', 22) * 16)
+    .duration(400).EUt(VA[IV]).buildAndRegister();
+
+// Calculation Processor
+mods.gregtech.circuit_assembler.recipeBuilder()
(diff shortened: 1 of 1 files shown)


v3ect0rgames (approved):
These recipes look very useful for late game quick ae2 production.

IntegerLimit:
Do we want to increase the cost of these recipes, especially energy wise? I feel like this is too close to the initial introduction of AE2 parts, and I believe later recipes would be better.

IntegerLimit (approved):

PR #776, opened 2024-06-20
Add Packaged Draconic

### LINKED ISSUE #741: Add PackagedDraconic

tracer4b:
Adds Packaged Draconic, a mod for PAuto support for DE. Closes #741.

Recipes for PD items were changed and will be trivial to craft immediately when DE Fusion is unlocked.

- [x] Add mod and recipes
- [x] Add/update quests

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -720,6 +720,11 @@
       "required": true
     },
     {
+      "projectID": 833931,
+      "fileID": 5299168,
+      "required": true
+    },
+    {
       "projectID": 845779,
       "fileID": 5006537,
       "required": true
--- a/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
+++ b/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
@@ -49644,7 +49644,7 @@
       "properties:10": {
         "betterquesting:10": {
           "autoclaim:1": 0,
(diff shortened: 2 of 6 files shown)


IntegerLimit (changes requested):
The description here seems to have just been removed, instead of being replaced by the new description (hence causing every test workflow to fail)

IntegerLimit (approved):

PR #619, opened 2024-02-13
Update to GT 2.8.* & Integrate Nomi Labs

IntegerLimit:
# PR to Update GregTech CEu to v2.8.*, and to integrate Nomi Labs (our coremod)

## This Branch's builds are now on Nightly!

[Builds](https://nightly.link/Nomi-CEu/Nomi-CEu/workflows/testbuildpack/dev%2Fgt-2.8?preview)

## Problems Solved
- Fixes Multi Searching in JEI (https://github.com/Nomi-CEu/Nomi-CEu/issues/519)
- Fixes Vanilla Ores in JEI (https://github.com/Nomi-CEu/Nomi-CEu/issues/271)
- Addresses Buckets showing up in JEI (https://github.com/Nomi-CEu/Nomi-CEu/issues/602)
- Specifies Default Fluids for All Fluid Conflicts (https://github.com/Nomi-CEu/Nomi-CEu/issues/602)
- Removes Bogo Feature in Bogo Sort
- Fixes Oxygen, Hydrogen and Nitrogen showing up as a null texture in JEI/HEI
- Fixed IV Quad/Nonuple Hatches & Buses being uncraftable (https://github.com/Nomi-CEu/Nomi-CEu/issues/550)
- Fixes the language override for Rhodium Plated Lumium-Palladium
- Fixes NBT Recipe Problems (https://github.com/Nomi-CEu/Nomi-CEu/issues/541, https://github.com/Nomi-CEu/Nomi-CEu/issues/518, https://github.com/Nomi-CEu/Nomi-CEu/issues/435, https://github.com/Nomi-CEu/Nomi-CEu/issues/663, ...)
- Fixes HM Assembly Line Recipe Duplicate (https://github.com/Nomi-CEu/Nomi-CEu/issues/517)
- Fixes Some HM Assembly Line Recipes not having Research (https://github.com/Nomi-CEu/Nomi-CEu/issues/517, ...)
- Fixes Recycling Recipes

## New Features
- Removes Anvil XP Scaling
- New ArchitectureCraft GUI for Sawbench & New Slopes (Comes from [GTNH's Fork](https://github.com/GTNewHorizons/ArchitectureCraft))
- Custom Items, Blocks, and Multis Improvements
- EnderIO Capacitor and Custom Capacitor Improvements
- Draconic Energy Storage Core Improvements
- Remapping and Removal of old and deprecated items (such as Content Tweaker Blaze Powder + Dark Red Coal)
[...]

PR #782, opened 2024-06-30
Change Void Oil Deposit Back to Raw Oil

IntegerLimit:
This reverts a change made in #619 (specifically commits [`be70198` (#619)](https://github.com/Nomi-CEu/Nomi-CEu/pull/619/commits/be7019840ca837a8c35d3f38c5258bfd7740d5dd) and [`4220f6c` (#619)](https://github.com/Nomi-CEu/Nomi-CEu/pull/619/commits/4220f6c185318d6632d35ba2ec18b21adafd7981)), where the Void Raw Oil Deposit was changed from Raw Oil to Oil.

Although this was intended initially to maintain the same resource production as before the GregTech update, this decreased the production of Naphtha, leading to Gasoline and HoG being even less viable than before.

## Ratio Changes:
Before: 50mb (Oil) -> 20mb (Sulfuric Naphtha), Ratio of 5:2
After: 100mb (Raw Oil) -> 150mb (Sulfuric Naphtha), Ratio of 2:3

All other ratios are not changed (as Deposit has also been changed to give x2 amount).

### DIFF
--- a/overrides/config/gregtech/worldgen/fluid/void/infinite_raw_oil_deposit.json
+++ b/overrides/config/gregtech/worldgen/fluid/void/infinite_raw_oil_deposit.json
@@ -1,9 +1,9 @@
 {
   "weight": 100,
-  "name": "Void Infinite Oil Deposit",
+  "name": "Void Infinite Raw Oil Deposit",
   "yield": {
-    "min": 40,
-    "max": 40
+    "min": 80,
+    "max": 80
   },
   "depletion": {
     "amount": 0,
@@ -13,5 +13,5 @@
   "dimension_filter": [
     "dimension_id:119"
   ],
-  "fluid": "oil"
+  "fluid": "oil_medium"
 }


v3ect0rgames (approved):

v3ect0rgames (changes requested):
Doesn't raw oil use 100mb to make 20 mb sulfuric naphtha and normal oil uses 50mb to make 20 mb of sulfuric naphtha. Wouldn't keeping it as oil speed up naphtha production

IntegerLimit:
I'm not sure which resource it was... can you please check in-game? Currently unable to check.

v3ect0rgames:
your code does make it raw oil, but oil gives the most napthta

v3ect0rgames:
so this change should not be made

tracer4b (approved):

v3ect0rgames (approved):
Realized I was on an older version. Should be added.

v3ect0rgames:
Github is bugged, I somehow have permission to revert this pr

Issue #770, opened 2024-06-17
Item Duplication with Extra Utilites 2's Enchanter


Fix3dll:
### Nomi CEu Version

1.7-alpha-4

### Launcher

Prism Launcher

### Environment

Multiplayer server with 3 GB RAM on Ubuntu 22

### Mode

Normal Mode

### What Happened

I can duplicate wirecutter (didnt tried others yet) with Extra Utilities 2's Enchanter.
![eu2's enchanter](https://github.com/Nomi-CEu/Nomi-CEu/assets/10743391/8c644fe5-592b-4131-b4e9-e8980f528a13)

### Expected Behavior

Expected to produce enchanted wirecutter by consuming old wirecutter

### Reproduction Steps

Try to enchant damaged steel wirecutter with 1 Lapis Lazuli on Extra Utilities 2's Enchanter.

v3ect0rgames:
For some reason the enchanter adds damage to the items instead of removing them

IntegerLimit:
I'll fix this via mixin in labs.

IntegerLimit:
Not closing this until Labs is updated in Nomi-CEu.

PR #786, opened 2024-07-02
Implement #770

v3ect0rgames:
Fix ae2 interface bug


v3ect0rgames:
something weird is going on, it wont let me base this branch off of nomi-ceu main

IntegerLimit:
Don't update AE2 yet, current update has a lot of issues. Leave this to me in combination with the UT + AE2 Stuff Updates.

PR #787, opened 2024-07-02
Implement #770

v3ect0rgames:
Fixes interface dupe


IntegerLimit:
We have been advised to skip the update of .55.6, and instead wait for the new release (due to many problems in .55.6 discovered by Exa in Nomifactory).

Closing this until a new AE2 update is pushed.

v3ect0rgames:
what about the beta version

PR #801, opened 2024-07-14
Fix AE2 Inscriber Recipes

HandsomeAsianMan:
This may need to be reverted if ae2 swaps .inscribe() and .press() back in a newer version than v0.56.5

### DIFF
--- a/overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
+++ b/overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
@@ -84,13 +84,13 @@ mods.appliedenergistics2.inscriber.recipeBuilder()
 	.input(ore('plateSilicon'))
 	.top(item('appliedenergistics2:material:19'))
 	.output(item('appliedenergistics2:material:20'))
-	.register()
+	.inscribe().register()
 
 mods.appliedenergistics2.inscriber.recipeBuilder()
 	.input(ore('itemSilicon'))
 	.top(item('appliedenergistics2:material:19'))
 	.output(item('appliedenergistics2:material:20'))
-	.register()
+	.inscribe().register()
 
 var applyProcessorChange = { OreDictIngredient lens, OreDictIngredient plate, ItemStack press, ItemStack circuit, ItemStack processor ->
 	/* Presses */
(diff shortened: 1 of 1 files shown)


v3ect0rgames (approved):

overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
```diff
@@ -117,7 +117,7 @@ var applyProcessorChange = { OreDictIngredient lens, OreDictIngredient plate, It
 		.top(circuit)
 		.bottom(item('appliedenergistics2:material:20')) // Silicon
 		.output(processor)
-		.inscribe().register()
```
IntegerLimit, line 120: Should this line have been changed?

IntegerLimit (changes requested):
Small comment, else looks good.

overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
HandsomeAsianMan: Yes, the inscribe and press behaviour has been swapped so keeping it as inscribed means the parts aren't used up

IntegerLimit: I see, thanks.

IntegerLimit (approved):

overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
IntegerLimit, line 120: Actually, one final change for clarity:

Can you add `.press` here?

IntegerLimit (changes requested):

overrides/groovy/postInit/Post-Initial/Main/Mod-Specific/ae2/misc.groovy
HandsomeAsianMan: Unfortunately mobile GitHub doesn't let me edit code so can't do that till tomorrow unless u can just take a sec to add that in there.

IntegerLimit (approved):

PR #806, opened 2024-07-15
Replace AE2 stuff -> AE2 stuff unofficial

HandsomeAsianMan:


IntegerLimit:
The Manifest Json must be sorted. Since this is such a small change, and this formatting change essentially takes the same time as just doing it separately, I will be closing this PR.

I will fix this in a PR in roughly 5 minutes.

PR #817, opened 2024-07-19
Make Transform QB Workflow use IDs

IntegerLimit:
This PR makes the output QB langs use the IDs of the quest line or the quest instead of their index, increasing consistency and removing the need for changes like https://github.com/Nomi-CEu/Nomi-CEu-Translations/pull/18 in the future.


IntegerLimit:
Because of the normalisation of IDs in the GT 2.8 QB PR, the only change to output langs as a result to this PR is a change in the lang key for the Matter Energy Quest Line. Therefore, translations will have to have an entry for both an index of 9 and an index of 10, for the Matter Energy Quest Line, until the next updat.e

PR #819, opened 2024-07-19
Update PackagedExCrafting and PackagedAuto

IntegerLimit:
This PR simply updates PackagedExCrafting and PackagedAuto, fixing game crashes and freezes when a PackagedExCrafting machine outputs an item.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -532,7 +532,7 @@
     },
     {
       "projectID": 308380,
-      "fileID": 5272321,
+      "fileID": 5521117,
       "required": true
     },
     {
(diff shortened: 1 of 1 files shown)


mvbarracuda2:
While the update to 1.7 beta5c fixes the crash when crafting with PackagedExCrafting and PackagedAuto, the crafted items aren't registered in the ME system. They just seem to vanish and the crafting job is still running without being ever finished.

I can provide an example savegame if that helps.

IntegerLimit:
Yes, that would be good. Thanks!

mvbarracuda2:
The issue was resolved by switching from ME import busses under the PackagedExCrafting crafting tables to EnderIO ME conduits. After that the items didn't vanish anymore, but were properly registered by the ME system.

Issue #271, opened 2023-03-03
Hide vanilla and non-obtainable ores


TheAirBlow:
They're just annoying when browsing through JEI, they always pop up first in the list. Additionally, hide overworld ore variant if it's and end ore for example.

IntegerLimit:
This will be part of a planned large cleanup of JEI/HEI.

PR #825, opened 2024-07-22
Finish #271

v3ect0rgames:


IntegerLimit (changes requested):
A lot of the GregTech items can be refactored into `metaitem`. Please take a look to see which ones can (most likely the ones with `gregtech:meta_` or `nomilabs:meta_`)

Note that for nomilabs meta items, whatever is outputted by `/gs hand` (say `metaitem('dustTinyDulysite')`), you will need to append `nomilabs` (e.g. `metaitem('nomilabs:dustTinyDulysite)`)

PR #843, opened 2024-07-31
Black Steel ABS Recipe

IntegerLimit:
This PR simply adds a Black Steel ABS recipe, which is the same as the original but with different inputs and fluid output amounts.

This PR also changes the shortcut Black Steel recipe to be 1.5x the duration of the normal recipe, instead of the previous 1.875x, following on from the original GT recipe.

Note that this PR is based off `labs-0.8.0` at the moment, as it requires changes made by that branch. Once that PR is merged, this PR will be rebased to be based off `main`. For now, note that https://github.com/Nomi-CEu/Nomi-CEu/commit/ae6f5684a5698545fe4dd71887bdc39b7f45acdb is the only commit that matters.


IntegerLimit:
Force Push was to re-base branch off labs update, which had a merge conflict fixed.

IntegerLimit:
Rebased off main.

PR #877, opened 2024-08-08
Reinstate that PAuto Packagers can be Placed Anywhere

Doniazade:
Adds one sentence to the pauto quest, see image. I haven't figured out the BQ scripts yet so you'll have to sync this to the other books manually as needed.

image

Tested on 1.7.6 nightly

(This information already exists in the extended crafting quest, but should be in the main pauto quest.)


IntegerLimit:
It would be greatly appreciated if the next PRs that you make be based from a branch that is not your 'main' branch.

IntegerLimit:
No need to sync to other quest books, I'll do that once this PR is done.

One change I would like you to make:
Can you change the 'Automating Extended Crafting' Quest so that, instead of saying 'place a Packager next to your Unpackager' it says 'place a Packager anywhere on your AE Network'?

Thanks in advance!

IntegerLimit (changes requested):
See Above.

IntegerLimit:
Another change: The text should not be green, as this is not a feature specific to Nomi-CEu.

Doniazade:
Will try to get these changes in tonight CET. Still learning github workflow.

Was unsure about the color scheme and what was notable positive vs ceu change.

IntegerLimit (approved):

PR #914, opened 2024-08-15
Make Tier IV Jetpack Hover Mode Stable

Doniazade:
Changes Vibrant and Resonant Jetpack (Vibrant being the relevant one) from very slow to zero gradual drop with Vibrant and Resonant jetpacks. This is to fix a major pain point in hard mode where no stable flight is not available until t4½mm because of Soul Binder>Angel Ring needing heads. No balance impact but major QoL when building, especially in the void.

Removes redundant status display of engine and hover status. Hover is obvious in use and engine on/off is not a functional keybind by default (G but doesn't work due to a conflict).

If you know what you're doing then Angel Ring is still a relevant goal because it can be stacked with a jetpack to increase speed.

### DIFF
--- a/overrides/config/simplyjetpacks-client.cfg
+++ b/overrides/config/simplyjetpacks-client.cfg
@@ -29,7 +29,7 @@
     B:"Enable Fuel HUD"=true
 
     # When enabled, a HUD that displays the states (engine/mode/etc.) of the currently worn jetpack or flux pack will show.
-    B:"Enable State HUD"=true
+    B:"Enable State HUD"=false
 
     # When enabled, switching jetpacks or flux packs on or off, or change their modes will display a status message above the inventory bar.
     B:"Enable State Messages"=true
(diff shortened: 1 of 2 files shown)


overrides/config/simplyjetpacks-client.cfg
```diff
@@ -29,7 +29,7 @@
     B:"Enable Fuel HUD"=true
 
     # When enabled, a HUD that displays the states (engine/mode/etc.) of the currently worn jetpack or flux pack will show.
-    B:"Enable State HUD"=true
+    B:"Enable State HUD"=false
```
IntegerLimit, line 32: I think that keeping the HUD would be good, just for consistency and info (for new players). I like the other changes (such as the HUD removing while chatting).

IntegerLimit (changes requested):

IntegerLimit (approved):

PR #917, opened 2024-08-16
Various QB Fixes and Cleanups

Doniazade:
Please double check so that the porting was done right, I think it was.

Changes dependencies in Mid Game to remove some unnecessary dependency lines. T3MM removed as platline dependency. First step of Mid Game rework.

Adds Nitric Acid optional dependencies on Ammonia and Nitrogen, since those are also valid Nitric routes. Unlinks Nitric visually from Nitrogen Dioxide. Adds optional dependency on Ender Air dist for Nitrogen Dioxide to hint that it's a good source later. (Nitrogen Dioxide quest could arguably be removed)

Hide SoC visual dependency on P boules since it's mostly irrelevant (the main gate is EV circ ass to use SoCs)

Removes mention of Platinum Hammer in Obsidian quest and rephrases the quest, mentions Void Crystals.

Adds mention of 25 home limit, mentions that enabling cheats disables /home countdown, gives example of /sethome with named home, mentions /setwarp as well as /warp quest. (This quest could be split into Lost Cities and Homes later)

Sets homes quest as another optional requirement for Interdimensional Travel, since it's the most common solution in use.

Splits RF+EU P2P off into separate quest, explains energy P2P. Adds note about shift scrolling P2P tunnels. (P2P quest is still quite long and Memory Cards could also be separated due to their use in cloning settings for other devices)

Screenshots of major changes:

image

image

image

image


v3ect0rgames (commented):
Looks good

Doniazade:
There is some discussion to be had about the t3mm dependency in platline. I'm gonna do some more adjustment there, basically the goal is to lead the player to t3mm for RTM, not platline. And sheldonite from t3mm is only one option for starting full platline, I prefer tetrahedrite in hm personally. I think the platline quest could be separated into Pt+Pd, Ru and rest.

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
```diff
@@ -10189,14 +10190,14 @@
...
           "partysinglereward:1": 0,
-          "questlogic:8": "AND",
+          "questlogic:8": "OR",
           "repeat_relative:1": 1,
           "repeattime:3": -1,
           "simultaneous:1": 0,
           "snd_complete:8": "minecraft:entity.player.levelup",
           "snd_update:8": "minecraft:entity.player.levelup",
           "tasklogic:8": "AND",
-          "visibility:8": "ALWAYS"
```
IntegerLimit, line 10199: Why did you hide Nitrogen Dioxide?

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
```diff
@@ -10243,13 +10246,13 @@
...
           "partysinglereward:1": 0,
-          "questlogic:8": "AND",
+          "questlogic:8": "OR",
           "repeat_relative:1": 1,
           "repeattime:3": -1,
           "simultaneous:1": 0,
           "snd_complete:8": "minecraft:entity.player.levelup",
           "snd_update:8": "minecraft:entity.player.levelup",
-          "tasklogic:8": "AND",
+          "tasklogic:8": "OR",
```
IntegerLimit, line 10255: Only one task, no need to make this `OR`

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
```diff
@@ -55623,6 +55640,65 @@
           "taskID:8": "bq_standard:retrieval"
         }
       }
+    },
```
IntegerLimit, line 55643: Note: Port QB doesn't add this quest visually to the quest book, only adds the quest in the database. It also doesn't rearrange any quests. Please import the HM qb to perform those changes.

IntegerLimit (commented):
Same questions for all QB files, but only pointed them out for one (for readability)

IntegerLimit (changes requested):
See Above.

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
Doniazade: > Why did you hide Nitrogen Dioxide?

The thought was that it's not required for anything in the tab, it's the most energy efficient option for Nitric Acid but the LCR recipe is also valid. I'll unhide it and just keep the dependency link to Nitric Acid as mouseover only, to reduce unnececssary lines.

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
Doniazade: Fixed, I think this was introduced because I initially didn't see the button for the requirement logic.

overrides/config-overrides/expert/betterquesting/DefaultQuests.json
IntegerLimit: Yep, makes sense, but since the quest still has things depending on it and it has a task, it won't work properly. Keep it unhidden for now, we can work out if we should remove it in the future.

Doniazade:
Doing the change in the HM book also introduces a slight change in the code of a couple other HM quests, I'm guessing this is due to disagreements between the scripts and bq? I don't know if these should be left in or removed.

IntegerLimit:
> Doing the change in the HM book also introduces a slight change in the code of a couple other HM quests, I'm guessing this is due to disagreements between the scripts and bq? I don't know if these should be left in or removed.

That is fine, and is due to ordering in JSON. The script may not order outputs correctly, but it is always fixed after importing, and scripts will respect new orders.

IntegerLimit (approved):

PR #928, opened 2024-08-20
Update Quest Book for AE2 & NAE2 Updates

IntegerLimit:
This PR adds the new AE2/NAE2 features to the quest book, as well as making minor improvements and touchups to AE2 quests.

Some New Quests were sourced from Nomifactory.


v3ect0rgames:
looks okay to me in terms of code errors

v3ect0rgames:
my game crashed once when loading this via curseforge but now it seems to be fine

v3ect0rgames:
what does linking quantum link chambers with quantum entangled singularity's do, don't you need a quantum link card

v3ect0rgames:
other than that clarity and gameplay also looks good.

v3ect0rgames (changes requested):
Clarify that formatting drives only reduces item types on the drive, and formatting adds more room to a specific item.

Doniazade (changes requested):
**I've reviewed this only the quests listed in the commits based on how the quests read in game based on the quests listed in the history, not in terms of code.**

Layout generally feels pretty good, it's very blocky but this is not something to fix now as mentioned.

Void Cell: Good warnings, great clarity. Really great that you explained the IO Port use case.

Matter Condenser: **Incomplete, you also need to drop 1 Enderpearl Dust with the singularity.** I verified that this is still the case.

Exposer: Great comparison to Controller Slaves, should be intuitive for many players. I like avoiding s's as in "Bus's" because it's slightly distracting, but that's only style. It's grammatically correct.

Beam Formers: **Maybe mention the Cleanroom use case** explicitly.

Interface P2P: Great explanation that it's like all the inputs are adjacent to all the outputs, that's my preferred explanation as well. However it should **also mention that the output side can pull from every input**, one use case I like is smart filter robot arms in my electrolyzers pulling from multiple stocked interfaces in the earlygame.

Programmed Circuit Card and Pattern Expansion Card: All good.

Crafting Card: Good job dispelling the performance myths when it comes to this. Concise summary. I was thinking about whether to include Export Bus here but I think it's an obsolete use case now, I think for pretty much every case stocking in an Interface is better due to the massive number of items you can set to be stocked.

Storage components: Good, I think this is an appropriate level of explanation of the 4 for 3 upgrade ratio vs fixed number of types tradeoff. As long as the player is aware of both they'll hopefully make good decisions.
**In my opinion the icon for Digital Storage At Last should be a 1k storage component** so that we have a complete set of both types of components.

Inscription Plates: **I don't think the requirements are appropriate** because the normal way to get the plates is to just find a couple of meteorites while exploring. You usually don't need the laser recipe or the compass. I think probably just unlock with Genesis and leave Inscriber as it is since it requires Energy Acceptor.

ME Interfaces: **Typo, "Locking mode" sixth paragraph**. The usual term for the non block form I see is slab form, although use your own judgement.

All quests not mentioned: All good/no comment.

v3ect0rgames (approved):

v3ect0rgames (changes requested):
Mention that fomatting allows more of that item in the cell, other then that looks good.

v3ect0rgames (approved):
Actually its fine.

IntegerLimit:
Ignoring comment about deps for Inscriber, most beginners do not find any inscriptions before getting the compass.

Doniazade (approved):
I'm satisfied with the changes, good to go.

PR #959, opened 2024-08-26
Add a FPS Impact Warning to EIO Facades

IntegerLimit:
This PR simply adds a tooltip to EIO facades, warning of FPS lag if used excessively.


MatthiasMann:
Huh - I always expected that non transparent facades (these which hide the conduits inside) actually improve the FPS

Folanlron:
The issue is that the YETA wrench..

Because there is no "check" for any kind of cooldown on the Facades going transparent, the facades will go "IS PLAYER HOLDING WRENCH", "NO THEN KEEP LOOPING".

When you get to about 200'ish Facades the problem starts,

PR #997, opened 2024-09-14
Questbook: Changed LuV Alloy quest Coil

smallming675:
This PR Changes the coil mentions in the Advanced LuV Alloy quest from Tungstensteel coil to RTM Alloy coil.


IntegerLimit (changes requested):
Port to HM.

PR #998, opened 2024-09-14
Questbook: added OR task for marked fusion injectors

smallming675:
Added an extra OR task for the Basic Fusion injector quest to include marked fusion injectors.


IntegerLimit (changes requested):
Run Port QB.

IntegerLimit (changes requested):
Also, task seems wrong. Shouldn't it be normal injector + core, then marked injector + marked core?

PR #1000, opened 2024-09-15
Questbook: Added quest for void satchel

smallming675:
This PR Added a quest for the Void Satchel Item, mentioning that it could auto pickup and void items.


IntegerLimit (changes requested):
Quest not actually visible in HM (needs to be added to questline, not just database)

Otherwise, looks good!

PR #1003, opened 2024-09-15
Questbook: Moved gas collectors and air centrifuge quests 

smallming675:
This PR Moves the Gas Collector and the Air Centrifuge Quest to The Beginning tab to EBF.


IntegerLimit (changes requested):
I would rather this be both in its current chapter, and in the beginning.

I also think it should be either an optional dependency to EBF or (quest wise) unlocked after the EBF, as it is not mandatory, and usually is not performed for first (few) EBF recipes.

PR #1004, opened 2024-09-15
Questbook: Merges the quests for Naquadah Reactor I and II

smallming675:
This PR Merges the Naquadah Reactor I and II quests into 1 quest as the Naquadah Reactor 1 aren't that useful.


IntegerLimit (changes requested):
Delete Old Naq Reactor Quests. Actually add the Quests to HM Questlines (currently only in database, not actually visible. Do through the edit button on left bottom.).

PR #1005, opened 2024-09-15
Questbook: Splitted "New Cable Coverings" into SBR and SiB.

smallming675:
This PR Splits the "New Cable Coverings" into Styrene-Butadiene Rubber and Silicone Rubber for better readability


IntegerLimit:
I don't think this is required, especially considering that both can be made when that quest is required, the main information is readable, and the quest is still relatively short.

PR #1038, opened 2024-10-07
Cleanup of Quest Book Layout

IntegerLimit:
This PR slightly cleans up the QB, adding information about NBT clearing, and cleaning up lines.


Doniazade (approved):
Reviewed in-game and looked through the code changes, looks good. Wasn't aware of the head swap feature for tools, definitely good to get that into the quest.

PR #1044, opened 2024-10-08
Migrating to AR reworked

ChromaPIE:
This PR migrates the old AR and LibVulpes to their forks (https://www.curseforge.com/minecraft/mc-mods/advanced-rocketry-2, https://www.curseforge.com/minecraft/mc-mods/library-for-ar-reworked). Config updated and double checked. Scripts updated but only to solve errors ingame, there might be some lines that are no longer needed in the forked AR but not removed.

- No error reported, doesn't seem to cause anything to malfunction, but I personally haven't had a chance to further explore the mod so there's still a chance it leads to, although not likely, some problems.
- Nomi Labs needs to be updated to remove codes related to solving the old bugs from the old AR that has been fixed in the forked version. Now with Nomi Labs 0.9.4 the game crashes on startup: https://mclo.gs/YKeFhvb.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -153,13 +153,13 @@
       "required": true
     },
     {
-      "projectID": 236541,
-      "fileID": 2949886,
+      "projectID": 1038780,
+      "fileID": 5732205,
       "required": true
     },
     {
-      "projectID": 236542,
-      "fileID": 3015079,
+      "projectID": 1024225,
+      "fileID": 5723687,
       "required": true
     },
     {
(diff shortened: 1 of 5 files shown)


IntegerLimit:
This will not be conducted until the mod has matured and stabilised. It is a very new mod, with a not very known developer, so we will wait for a period of time. We'll see how this one turns out.

I'm going to close this PR for now. Can you instead make an issue, posting the link to the two mods, so we remember and can reconsider in the future?

ChromaPIE:
> This will not be conducted until the mod has matured and stabilised. It is a very new mod, with a not very known developer, so we will wait for a period of time. We'll see how this one turns out.
> 
> I'm going to close this PR for now. Can you instead make an issue, posting the link to the two mods, so we remember and can reconsider in the future?

Understandable. Issue opened at https://github.com/Nomi-CEu/Nomi-CEu/issues/1045

PR #1065, opened 2024-10-14
Hide Higher Tier Muffler Hatches

IntegerLimit:
This PR simply hides higher tier muffler hatches (anything higher than LV). They tend to simply confuse players.


v3ect0rgames (approved):

tracer4b:
Maybe consider renaming the remaining muffler hatch to simply "Muffler Hatch" then?

IntegerLimit:
> Maybe consider renaming the remaining muffler hatch to simply "Muffler Hatch" then?
> 
> Higher tier muffler hatches do have a use in producing more ash byproduct, even if hardly anyone uses that properly, it is a feature we are "removing"

Also, ash byproduct can't even be removed from the hatches with automation, so I'm not sure it has a purpose at all.

Renaming sounds like a good idea.

PR #1069, opened 2024-10-14
Misc Questbook Tweaks

smallming675:
This PR Implements serveral misc. questbook fixes:

- [x] Move Stainless Steel down so it doesn’t look like it’s a requirement for The Second Tier One Circuits. Make it a right angle against Distillation Tower
- [x] Move PVC out of line with PE/PVC,
- [x] Move Persulfate & Iron III Chloride quest next to Plastic Substrate and link it. The LV Chemical Reactor requirement is largely irrelevant. Remove the LV Chemical Reactor quest from Early Game
- [x] Remove visual link between HV Assembler and HV Energy Hatch. If you have a Cleanroom then you almost certainly have a HV Assember. Cleanroom is the main requirement
- [x] The horizontal line containing PE, PVC, plastic substrate is quite confusing. Overlapping dependency lines.
- [x] Unlink Circuit Wafers from MV engraver. It is already required via Lenses
- [x] Unlink Distillery visually from Ethylene, since it’s already linked via Navigating Petrochem
- [x] Move Plastic Substrate out of the vertical line connecting HV Hull and Cleanroom. It’s not connected to anything along that line. 
- [x] Consider unlinking Kanthal from Chrome, or rearranging to avoiding that horizontal line overlapping Circuit Wafers and Lenses. Those two quests could perhaps be moved up out of that line
- [x] In The Beginning, unlink LV Electrolyzer from LV Wiremill. Not required, cuts across another line
- [x] Check completeness of Discord quest to make sure it has all the useful information. Especially %t tags and %t searching
- [x] Fixed several typos

(ps sry for the late notice, i have some other work to do)


IntegerLimit:
This looks good in terms of the changes, but I do want to quickly check quest layouts in game.

IntegerLimit (changes requested):
Can you, in the quest that mentions FTB Homes and Warps (in Genesis I think), that warps can only be created and removed by people with operator permissions?

IntegerLimit (approved):

PR #1079, opened 2024-10-17
Add Tiny Coke

IntegerLimit:
This PR adds Tiny Coke, which allows for players to turn their Coke into fuels that are not wasted in the Furnace. More helpful for Hard Mode players.

Also improves the quest book to highlight tiny fuels more.


IntegerLimit:
Force Push was to fix merge conflicts.

PR #1108, opened 2024-11-18
Bloat removal

D-Alessian:
These mods should be useless. Maybe there are more


D-Alessian:
why are all the merges there as commits?

PR #1119, opened 2024-11-23
Questbook fixes

smallming675:
This PR fixed breaking changes within the questbook, including:
- Palladium quest now requires platline quests so that it is completable
- Gasoline quest now requires Navigate Petrochem as the previous prerequisites prerequisites were deleted


smallming675:
Whoops didn't meant that

PR #1120, opened 2024-11-23
Fix Uncompletable Quests

smallming675:
This PR fixed breaking changes within the questbook, including:
- Palladium quest now requires platline quests so that it is copmletable
- Gasoline quest now requires Navigate Petrochem as the previous prerequisites prerequisites were deleted


IntegerLimit:
Needed in HM?

IntegerLimit:
Note: This is not required in HM.

IntegerLimit (approved):

PR #1134, opened 2024-12-05
Further Standardize Formatting in Quest Books

IntegerLimit:
This PR has five changes, made to the relevant check to `checkQB` in all quest book formattable fields, including quest names, quest descriptions, quest line names, and quest line descriptions.

In addition to the five practical changes, this PR also greatly reformats the quest book checking scripts.

## 'Double' Formatting
This PR removes 'double' formatting, which occurs when a formatting code is used, then some text was inputted, and then the same formatting code was used, resulting in duplicate and redundant formatting.

This has no effect on the QB in game.

This issue was first reported by @marisathewitch.

## Double Spaces
This PR removes all double spaces, which are unintentional and are hard to catch outside of automated scripts.

This does have a positive effect on the QB in game.

## Formatting Order
This PR standardises formatting signals, with their position, in respect to spaces or newlines around it:

Non 'Resetting' signals: to be after any spaces
'Resetting' signals: to be before any spaces

This cleans up the QB JSON, and makes it easier to read. This has no effect in-game, apart from some end of line spaces being removed as a result.

## Formatting At End of String
This PR removes any redundant formatting signals at the end of the string, as they are irrelevant and simply clutter the strings.

However, if a name or description changes the formatting signal, and the signal is not reset by the end of the string, this PR adds a resetting formatting signal, to reduce confusion in future edits.

This has no effect on the QB in game.

## Expansion of Redundant Formatting Checking Across Newlines
An older PR, #962, made checking of redundant formatting ignore spaces between formatting signals. This PR expands it out to include newlines, further removing unneeded signals.


IntegerLimit:
Force push was due to incorrect removal of 'double' formatting.

IntegerLimit:
Force Push was due to missing block in previous commit.

v3ect0rgames (approved):
Don't see any code errors, just make sure its tested in game.

PR #1139, opened 2024-12-08
Small Earlygame Tweaks

IntegerLimit:
This PR makes some small earlygame tweaks to expert mode:

- Adds a Log -> Stick Shortcut Recipe
- Removes Shortcut Paper Recipes
- Makes Wood Pulp Output 2 Instead of 4

### DIFF
--- a/overrides/groovy/postInit/Post-Initial/Main/General/Early-Game/earlygame.groovy
+++ b/overrides/groovy/postInit/Post-Initial/Main/General/Early-Game/earlygame.groovy
@@ -9,25 +9,6 @@ import net.minecraftforge.fluids.FluidStack
 import static com.nomiceu.nomilabs.groovy.GroovyHelpers.RecyclingHelpers.*
 import static gregtech.api.GTValues.*
 
-// Allow treated wood sticks in (most) stick recipes
-ore('stickWood').add(metaitem('stickTreatedWood'))
-
-// Make Wood Frame Box only accept non-treated wood sticks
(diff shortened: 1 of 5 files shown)


overrides/groovy/postInit/Post-Initial/Main/General/Early-Game/earlygame.groovy
```diff
@@ -9,6 +9,58 @@ import net.minecraftforge.fluids.FluidStack
 import static com.nomiceu.nomilabs.groovy.GroovyHelpers.RecyclingHelpers.*
 import static gregtech.api.GTValues.*
 
+// Allow treated wood sticks in (most) stick recipes
+ore('stickWood').add(metaitem('stickTreatedWood'))
```
tracer4b, line 13: By right this shouldn't be allowed, treated wood is a separate material from normal wood (similar to iron/wrought iron).
If the players think this is a reasonable QoL it can go in though

tracer4b (changes requested):

overrides/groovy/postInit/Post-Initial/Main/General/Early-Game/earlygame.groovy
ProTriforcer: I think this will create more annoyances than it solves. Creosote is much less free than wood and you might accidentally waste treated sticks with jei's + button

tracer4b (approved):

PR #1161, opened 2024-12-23
Fix Tools Compat with Node 22

IntegerLimit:
This PR removes the usage of 'input assertions', which are removed in Node 22.

See [Stack Overflow Post](https://stackoverflow.com/questions/78876691/syntaxerror-unexpected-identifier-assert-on-json-import-in-node-v22).


IntegerLimit:
Note that tools now only works for Node 18+.

PR #1166, opened 2024-12-29
Change Actually Additions' Laser Tool to GT Screwdriver

ChromaPIE:
This PR is to change the tool used to configure Actually Additions' laser relays (switching I/O mode for Fluid relay and for Item relay, configure priority) from vanilla Compass to GregTech Screwdriver, so players don't have to carry an extra tool with them.

It contains a simple change to AA's config file and an extra line in `resources/gregtech/lang/en_us.lang` to prevent the tip text shown when looking at a laser while holding the Laser Wrench from showing the raw localization key `item.gt.tool.screwdriver.name.name`.

Tested and functioning as intended.


IntegerLimit (approved):

IntegerLimit:
Thank you for the PR! Changes looks good, and should be a worthwhile change.

PR #1170, opened 2024-12-31
Update Build Pack Lang Job Name

IntegerLimit:
This PR simply changes the Build Pack Lang job to reflect the fact that it no longer builds, or uploads, the changelog, as of #1023.


IntegerLimit:
Is this just an excuse so I can fix my previous merge's commit syntax? Perhaps.

PR #1175, opened 2025-01-01
Fluid Ingredient JEI Improvements

IntegerLimit:
This PR makes the following changes to how wild fluid containers appear in JEI:
- The ingredient no longer cycles, only displaying buckets instead
  - Firebricks recipe displays cell
  - Reduces confusion for crafts, and tooltips + QB already mention the acceptance of any fluid container
- Fixes localization with GT material fluids in Input Tooltip


v3ect0rgames (approved):
Don't see any code errors. Looks Good

PR #1184, opened 2025-01-04
Add Tritanium Recipe to Universal Crystalizer

IntegerLimit:
This PR adds a recipe for Tritanium in the Universal Crystalizer.

The recipe costs 2 titanium and 1 duranium for each tritanium, unlike the fusion recipe, which costs 3 titanium and 2 duranium for each tritanium. This should help reduce duranium costs. At a base power level, excluding parallels and OCs (which can be done a lot on the Crystalizer, due to Laser Hatches), the recipe is 21.6x faster than a Fusion MK III.

It also shortens durations (of DE crafts), slightly tweaks energy consumptions (shouldn't really matter, players should have source anyways), and moves the recipes to GroovyScript.


v3ect0rgames (approved):

v3ect0rgames:
ok undid it

Issue #1178, opened 2025-01-02
Improve NC fission for the time being


D-Alessian:
### Nomi CEu Version

Nightly

### Launcher

N/A

### Configurations or Mods Changed

N/A

### Environment

N/A

### Mode

Both Modes

### Related Problem

NC fission currently has a LOT of problems. I'll try to list them here.

1. The setups being complex to build means players NEED to use building gadgets for big setups, but this has a couple problems:

 - most players don't know they can transform and use BG strings in game to insta build their reactors
 - you need to use GregTech graphite blocks in CEu since they're the only ones craftable, but both leu-235.com and hellrage's Reactor Planner use NC graphite blocks by default. It is possible to transform the string to a usable one but it's too annoying for the regular players to do, so they're forced to use existing BG strings
 - it can't place down active coolers
 
2. You just need too much centrifuging for it to be worth it. You either need to place down 400 HV centrifuges if you want optimal power ratio, or place one GCYM that will eat up half the power you'll make out of the reactor eventually, or more for bigger setups. 

3. It needs a lot of tungsten to make the casings and that deters a lot of players from actually building the reactors.

4. TBU is the only fuel that is even remotely viable for powergen.

5. Cryotheum is the only coolant remotely viable for cooling, and it's not passivable in HM, making actively cooled reactors basically impossible.

6. They are really big, which causes people to be unsure of what is what when looking at schematics, but also can get visually unappealing and makes it hard to build them by hand.

### Your Solution

Solution 1 : Buff U-235 from red granite

Solution 2 : Buff coolers that aren't cryotheum

Solution 3 : Implement a kovarex enrichment process, like in factorio, to get U-235

Solution 4 : Make the reactors more compact, improve the power ratio, make them consume fuel faster, but increase the cost of everything as well to compensate.

### How Will Your Solution Work

Solution 1 :  Makes LEU a viable fuel, and reduces the overall size needed
[Easy to implement, Boring Gameplay] [Fixes 2, 4] [Helps with 1, 3, 6]

Solution 2: Better coolers would mean more interesting and complex reactor designs, it would mean better power, smaller reactors needed, and using coolants like helium would make it possible to use active cooling in HM
[Easy to implement, Somewhat Interesting gameplay] [Fixes 2, 5] [Helps with 1, 3, 6]

Solution 3 : A better solution than solution 1 to the whole U-235 problem, a nice nod to factorio and a way to implement some interesting line with retroaction.
[...]

PR #1188, opened 2025-01-07
Misc NuclearCraft Balancing Tweaks

IntegerLimit:
This PR makes several changes to make NuclearCraft. These are aimed at several prospects, but mainly, it aims to improve prospects for NuclearCraft power generation, especially in Hard Mode, and aims to make NuclearCraft easier to manage and more interesting in the endgame, especially for Hard Mode.

This is done through 3 changes:
1. Uranium 238 from Renewable Red Granite has been buffed to 1% (+1%), making Uranium a viable fuel source for passive breeding and power generation (when combined with Uranium 233 from TBU Breeding)
2. Thermal Centrifuge Times for Prepared Thorium have been reduced by ~14%, to reduce setup and power consumption of a passive TBU reactor
3. Yields from Depleted Fuels have been increased by 10%, to increase production of NC isotopes for HM endgame

This PR also makes two misc changes:
1. Removes the Thermal Centrifuge Recipe for NC U238. GT U238 can do everything the NC version can, and the NC version is hidden from JEI
2. Removes recipes for NC U238 Block, and hides it from JEI. Testing showed that the recipe could not be used (due to a conflict with a GT recipe), thus making the block unobtainable, but a conversion recipe to the GT U238 Block has been added, just in case.

Closes https://github.com/Nomi-CEu/Nomi-CEu/issues/1178


v3ect0rgames (approved):
Looks good

PR #1189, opened 2025-01-08
Disable All Draconic Evolution World Generation

officereso:
added valid dimensions from /advRocketry planet list to oreGenDimentionBlacklist


IntegerLimit (commented):
Question: instead of doing this, could we just completely disable Draconic Evolution's World Generation (setting below)? Can you do some testing with that setting enabled?

officereso:
Setting B:worldGenEnabled to false does indeed fix the issue of Dracoinum spawning on other planets.

v3ect0rgames (approved):

PR #1209, opened 2025-01-18
Fix fluidCrafting.groovy replacing shapeless cable recipes

officereso:
We're already removing the recipe using ``crafting.remove``. ``.replace()`` is removing the shapeless recipes for cable_dense_covered, cable_smart, and cable_dense_smart.


IntegerLimit (approved):
Thank you.

PR #1215, opened 2025-01-29
Fix Occasional XP Disappearances when Dim Changing

IntegerLimit:
This PR fixes xp sometimes disappearing when changing dimensions.

Whilst this bug has been mostly fixed by Universal Tweaks, there have been reports that this can still sometimes occur, but adding this mod prevents all occurances.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -593,6 +593,11 @@
       "required": true
     },
     {
+      "projectID": 486778,
+      "fileID": 3372374,
+      "required": true
+    },
+    {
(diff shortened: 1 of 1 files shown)


IntegerLimit:
Note: this mod has third party sharing disabled. I will contact the mod author first, but failing that, I will implement this into Labs (the mod is MIT, the change is tiny).

IntegerLimit:
The mod has enabled third party downloads at my request.

PR #1250, opened 2025-03-20
Fix Various Typos

D-Alessian:
Fixes the issue mentionned in Nomi-CEu Development considering the typo and the syntax error in the AA tooltip. I also took the liberty to use typo finding tools to fix various typos. This doesn't fix them ALL as tools that could indentify them all would lead to way too many false positives, taking an absurd amount of time to check all of them. 

I used https://github.com/crate-ci/typos and https://github.com/codespell-project/codespell , but checked every change made manually. The only one that changes to variable names I did was in the _oreDict.zs CT script because cmon, it's ugly. 

There are a LOT more typos but a lot of them are due to other mods, either them making typos in item names or in the mod's config, hence why I ignored all the configs except QB.


IntegerLimit:
Can we add the spell check to a CI script?

D-Alessian:
> Can we add the spell check to a CI script? (That would mean you add the toml back, and we do some other stuff)
> 
> I'll investigate this shortly.

I mean, yeah, could be possible. But would require probably more work to add all the exceptions for the false positives it gives. There were a LOT of them. And there doesn't seem to be, from what I've tried, a nice regex to catch all of them, it's really a case by case thing.

Or maybe, run the action only on the PR's content, and have it only detect typos, but not block the merge?

IntegerLimit:
I'll work on it later. Let's merge this first.

IntegerLimit (approved):
LGTM.

IntegerLimit:
Can you post the toml config file you used here? Just as a starting point.

D-Alessian:
> Can you post the toml config file you used here? Just as a starting point.

I haven't kept it, it should still be in the commit but it's very simple. Just excludes the config and config-overrides folders and a basic regex that didn't work like I thought it would (I thought I could tell it that anything starting with § using /§\w+/ should be ignored, but that's not how the regex exclusion works. It excludes it after finding the typos, not before searching for them)

D-Alessian:
```
[files]
extend-exclude = ["overrides/config/**", "overrides/config-overrides/**"]

[default]
extend-ignore-words-re = ["§\\w+"]
```

IntegerLimit:
Right... I should have looked at the commit log. Thanks.

PR #1253, opened 2025-03-22
Deprecate the AA Player Interface

### LINKED ISSUE #1180: Hide AA Player Interface and use Player Chest instead
### Nomi CEu Version

Nightly

### Launcher

Prism

_No response_

### Environment

Singleplayer

### Mode

Normal Mode

### Related Problem

**What is Player Interface/Chest?**
A block that simulates the inventory of the player who places it for automated IO. A use case would be automatically extract empty cup from player's inventory, send it to the AA coffee maker and deliver the brewed coffee straight to player.

**Differences between the two?**
Player Chest from Extra Utilities 2 allows for per-slot configuration. You can configure No access/Full access and I/O only for each inventory/armor/hotbar slot, while for AA Player Interface, it's not configurable at all.

**Problem with the AA one?**
Upon placed, EnderIO conduit will NOT recognize it as a valid container, players will have to break and place the conduit again for the connection. What's more annoying is that, the disconnection happens every time on entering world, causing players to have to do it from time to time. the ExU2 one functions seamlessly, just like a normal chest.
Also, due to the design of Nomi, Player Interface got gated behind the Empowerer (Tungsten Steel), which hardly makes any sense for a fun gadget like that, while to make Player Chest you only need the Resonator.

### Your Solution

- Hide the AA Player Interface, remove the recipe
- Add a tooltip for it: Deprecated, known to have problem, use the ExU2 Player Chest, put it in the crafting grid to convert
- Add the conversion recipe

-

_No response_

### Additional Information

_No response_

D-Alessian:
Fixes #1180
Tested in nightly. 

Very straightforward. Only 2 lines of code.

### DIFF
--- a/overrides/groovy/postInit/main/general/misc/deprecation.groovy
+++ b/overrides/groovy/postInit/main/general/misc/deprecation.groovy
@@ -91,6 +91,10 @@ conversion(metaitem('nomilabs:plateMithril'), item('thermalfoundation:material',
 conversion(metaitem('nomilabs:ingotInfinity'), item('avaritia:resource', 6))
 conversion(metaitem('nomilabs:plateInfinity'), item('moreplates:infinity_plate'))
 
+// Deprecate Player Interface in Favor for the XU2 Player Chest
+deprecate(item('actuallyadditions:block_player_interface'))
+conversion(item('actuallyadditions:block_player_interface'), item('extrautils2:playerchest'))
+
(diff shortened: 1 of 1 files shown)


IntegerLimit:
I'll probably add a quest for this too, but that can go into `qb-misc-2`.

IntegerLimit (approved):
LGTM.

PR #1255, opened 2025-03-23
Add Typo Checking CI Tasks

IntegerLimit:
This PR is heavily inspired by the great efforts in #1250, and attempts to pull together a *somewhat* working system, that is reliable, to checking for typos.

Like #1250, this system uses https://github.com/crate-ci/typos. This system runs on the `checks` workflow task.

Checks are limited to lang files (including extracted QB titles and descriptions), with lang files having the formatting from them stripped (including substitution areas like %1), and files in the `.github` folder. Whilst #1250 did contain some typo fixes for code, this is not included in this PR.

Result:
Hopefully a large reduction in typos, especially for the quest book.

### DIFF
--- a/.github/workflows/buildpack.yml
+++ b/.github/workflows/buildpack.yml
@@ -30,11 +30,11 @@ on:
         required: false
         type: string
       compare_tag:
-        description: Tag(s) to compare against. If specifying multiple, seperate by commas. (Spaces allowed). See CONTRIBUTING.md for more information.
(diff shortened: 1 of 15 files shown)


IntegerLimit:
Note: This is reliable (in testing), and has minimal false positives. However, this does not allow for auto-fix, and does not have easy to read error messages, but it is a good warning system to have for now.

PR #1275, opened 2025-03-29
Update manifest.json

v3ect0rgames:


IntegerLimit (changes requested):
Thank you for the PR! Can you also update Had Enough Items?

v3ect0rgames:
Oh i thought we weren't sure if that would work.

IntegerLimit (approved):
LGTM. I've tested HEI update, it seems to work well.

PR #1337, opened 2025-05-20
Fix Fueling Station in Edge Case Environments

CiangCing14:


CiangCing14:
The rest bugs. I think we should ban the access right to land on gas giant. If you try to land on it, then you will achieve an infinite void.
The last bug is Space Station will disappear after save in Space Station and get in game world.

PR #1345, opened 2025-05-31
Buff Recipes for Red Alloy with Annealed Copper

IntegerLimit:
This PR buffs recipes for red alloy that use annealed copper, to reduce annoyance of many stacks of redstone being needed to make a small amount of alloy. The effect on balancing is minimal.

Alloy Smelter/Mixer: Redstone cost cut by half, annealed to redstone ratio is 1:2, compared with copper to redstone ratio of 1:4
ABS (NM): Annealed to redstone ratio is 1:1, compared with copper to redstone ratio of 2:3
ABS (HM): Same modifications as Alloy Smelter/Mixer

The decomposition recipes for both modes have been updated to reflect the cheapest recipe. These red alloy recipe changes have also been moved from `blastFurnace.groovy` to `materialChanges.groovy`.

### DIFF
--- a/overrides/groovy/postInit/main/general/misc/materialChanges.groovy
+++ b/overrides/groovy/postInit/main/general/misc/materialChanges.groovy
@@ -1,6 +1,7 @@
 package postInit.main.general.misc
 
 import com.nomiceu.nomilabs.groovy.ChangeRecipeBuilder
+import com.nomiceu.nomilabs.util.LabsModeHelper
 import gregtech.api.recipes.RecipeBuilder
 import gregtech.api.recipes.ingredients.GTRecipeInput
 import net.minecraftforge.fluids.FluidStack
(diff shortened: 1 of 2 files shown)


overrides/groovy/postInit/main/general/misc/materialChanges.groovy
```diff
@@ -78,3 +79,78 @@ material('black_steel')
...
+					recipe.clearInputs()
+						.inputs(ore('dustCopper') * 2, item('minecraft:redstone') * 3)
+						.circuitMeta(2)
+				}.replaceAndRegister()
+
+			builder.copyOriginal().clearCircuitMeta()
+				.changeEachFluidOutput { fluid -> fluid * (L * 2) }
+				.builder { RecipeBuilder recipe ->
+					recipe.clearInputs()
+						.inputs(ore('dustAnnealedCopper') * 2, item('minecraft:redstone') * 2)
```
D-Alessian, line 125: Why `2 + 2 -> 288` and not `1 + 1 -> 144`?

D-Alessian (commented):
I like this change overall. Giving ABS better yields is always nice too, it's heavily underutilized. 

One other note is that the formula is still `copper, 1, redstone, 4`, that should probably be changed. I know you made groovy methods specifically for that so it shouldn't be hard to do it here afaik?

doniascion:
It does affect redstone usage pretty significantly, I tried changing my patterns and a heart craft went from 68k to 49k redstone, using the 1:2 HM recipes. Basically for that craft red alloy accounted for over half of the redstone demand.

Together with the changes to *theum redstone cost the need for redstone has gone down a lot and I think that's a good thing overall, although I think redstone demand already went down a lot with the changes to the thermal powder/*theum recipes. Overall I like it since spamming T4 missions and processing a gorillion crushed redstone gets a bit much, especially before good oreproc infra.

It's a little weird that there would be a difference between annealed and regular copper for liquid recipes since it's just the same material except heat treated, but I think that can just be ignored.

PR #1349, opened 2025-06-03
Update ZBGT Addon script

Mazlaci:
Expanded version of the ZBGT addon script, implementing missing crafts with general cleanup

### JEI Cleanup
Removed Creative Item Bus from JEI, since the craft was already removed
Removed items without crafts and uses:
- Compact Fusion Coil MK-II
- Component Assembly Line Casing
- Cooland and SP Cells

### Wrap of _\#_
Removed crafts for Wraps with no uses, and hidden them from JEI
Exception are the lower tier Wrap of Circuite Boards and Wrap of Simple SoC, since higher tiers have viable uses, so it's better to keep them in line.

Added "unwrap" recipes for each Wrap in the assembler giving back the original 16 components

### Cells
Having no other uses I removed the smaller than 60K cells and the crafts for filling them.
The 60K cell already had an assembler recipe from normal gregtech empty cells, so I left only that one in.

### Dropper Cover
Added Assembler recipes for dropper cover LV to IV which previously only had normal crafts requiring a wrench.
The assembler recipes should be easier to automate with AE2.

### YOTTank
Previously every part of the YOTTank were craftable, exept the fluid Cell Blocks themself, wich meant the multiblock were unusable.

I have given an assembler craft for Fluid Cell Blocks T1-3. I balanced those around the single block tanks:
- Fluid Cell T1 &harr; Super Tank I
- Fluid Cell T2 &harr; Super Tank V
- Fluid Cell T3 &harr; Quantum Tank V

The tier 4-10 fluid cells were hidden from jei.

My philosopy was, T2-3 already give a quite big amout of fluid capacity, and if needing more, you can alwasy make more levels. for the multiblock.
This makes them inline with single-block tanks, without making them obsolite, and having a more interesting upgrade curve than just crafting bigger capacity,

It all makes the YOTTank quite balanced in my opinion. You can make a useful YOTTank at around from IV tech, meaning late midgame to early late-game

#### Bigger Parallel Hatches
I considered hiding and removing the recipe for higher than OpV Parallel Hatch (×65536 parallel), since these require voltage tiers available only with creative power in this pack.
But since there is a way for creative power i left them in.

#### Creative computing hatch, Creative data hatch
There is an active TODO in the file to implement the Creative computing hatch and Creative data hatch
I don't feel comfortable enough to give these balanced recipes, since I never played in survival until creative.
I did't hide them in JEI, since I feel like these could be part of the script at a different later PR

### DIFF
--- a/overrides/groovy/postInit/addon/zbgt.groovy
+++ b/overrides/groovy/postInit/addon/zbgt.groovy
@@ -21,6 +21,7 @@ crafting.remove('zbgt:creative_energy_hatch_to_emitter')
 // Creative Item Bus (Useless since you already have chest)
 crafting.remove('zbgt:creative_item_bus_to_chest')
 crafting.remove('zbgt:creative_chest_to_item_bus')
+mods.jei.ingredient.removeAndHide(item('gregtech:machine', 18008)) // Creative Item Bus
 
 // Creative Tank
 crafting.remove('zbgt:creative_fluid_hatch_to_tank')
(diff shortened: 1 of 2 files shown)


overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -21,6 +21,7 @@ crafting.remove('zbgt:creative_energy_hatch_to_emitter')
 // Creative Item Bus (Useless since you already have chest)
 crafting.remove('zbgt:creative_item_bus_to_chest')
 crafting.remove('zbgt:creative_chest_to_item_bus')
+mods.jei.ingredient.removeAndHide(item('gregtech:machine', 18008)) // Creative Item Bus
```
D-Alessian: don't remove creative items from JEI, as they can be used, well, for creative

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -221,3 +222,410 @@ if (LabsModeHelper.normal) {
...
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 6))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 7))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 8))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 9))
+
+
+/* Wraps */
+// Remove unused wrap crafts and from JEI
+// Wrap of Circuite board and SoC not removed, since other tiers have uses
+mods.gregtech.assembler.removeByOutput([item('zbgt:zbgt_meta_item', 57)], null) // Wrap of ULV Circuit
```
D-Alessian: You should use metaitem instead of item. You can get the metaitem of an item by running `/gs hand` with the item in your hand. Here it would be `metaitem('wrapped.circuit_board.extreme')`

Mazlaci:
Originally I felt like, having a bunch of unusable items in JEI mean clutter, and if anything confusing for someone new wanting to use the mod here.

But being version independent from the mod is a huge factor to consider for sure.

I still think a clean JEI is preferred, but I can readd them if that's better as an add-on mod.

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -221,3 +222,410 @@ if (LabsModeHelper.normal) {
...
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 5))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 6))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 7))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 8))
+mods.jei.ingredient.removeAndHide(item('zbgt:yottank_cell', 9))
+
+
+/* Wraps */
+// Remove unused wrap crafts and from JEI
+// Wrap of Circuite board and SoC not removed, since other tiers have uses
```
D-Alessian: here, considering the amount of items removed and how repetitive it can be, by also having the recipe to unwrap them, it would probably better to have an array of the common part of the metanames between wrapped and unwrapped stuff, of all the items used, and just call both functions on it. For the recipe you could, for SMDs for example.

```
mods.gregtech.assembler.recipeBuilder()
	.inputs(´metaitem('wrapped.' + metaname, 72))
	.circuitMeta(1)
	.outputs(metaitem('component.' + metaname) * 16)
	.duration(100).EUt(VA[LV])
	.buildAndRegister()
```
	
Just a thought here. 
But yeah, seems like iteration would make the code cleaner. Probably with a map of the base name that's in common and the (smd.capacitor) and the part that isn't for the unwrapped version (component.). 

Up to you really, just thought I'd share.

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -221,3 +222,410 @@ if (LabsModeHelper.normal) {
...
+	)
+	.outputs(item('zbgt:zbgt_meta_item', 121)) // IV Dropper Cover
+	.duration(200).EUt(VA[IV])
+	.buildAndRegister()
+
+// Remove other unused blocks from JEI
+mods.jei.ingredient.removeAndHide(item('zbgt:misc_casing', 4)) // Compact Fusion Coil MK-II Prototype
+mods.jei.ingredient.removeAndHide(item('zbgt:misc_casing', 5)) // Compact Fusion Coil MK-II Finaltype
+mods.jei.ingredient.removeAndHide(item('zbgt:coal_casing', 9)) // Component Assembly Line Casing (UHV)
+mods.jei.ingredient.removeAndHide(item('zbgt:coal_casing', 10)) // Component Assembly Line Casing (UEV)
```
D-Alessian, line 424: Those for UEV and up are 100% safe to remove, even considering it might update, since UEV+ won't come into play in Nomi-CEu (without a change to core progression).

D-Alessian (changes requested):
Overall I love the fact you started working on the ZBGT compat too, my original compat is too light, and it def needs some polishing. 

Just yeah, the metaitem is the biggest part, it's gonna be annoying changing 700 trillion items, but it really is the standard in all other situations, and I already know Int is going to request that before this is allowed to be pushed. Iteration could make some parts of the code cleaner, mainly the big block from 271-532. 

And last part of the feedback is what I said before, perhaps hiding some of these would mean requiring a re-review later and making this script version dependent. Not saying it needs to go, just saying we need to think about it (ultimately down to Int).

But like yeah, ultimately I do love that you got on to improving the compat!

D-Alessian:
> Originally I felt like, having a bunch of unusable items in JEI mean clutter, and if anything confusing for someone new wanting to use the mod here.
> 
> But being version independent from the mod is a huge factor to consider for sure.
> 
> I still think a clean JEI is preferred, but I can readd them if that's better as an add-on mod.

I think waiting for Int's opinion on this is the best move. There's pros and cons. It can also be quite easy to edit the file when Zorbatron DOES make some of it usable since I follow ZBGT closely and speak to Zorb daily. So it's not like ANY mod that we won't follow up on

IntegerLimit:
> > Originally I felt like, having a bunch of unusable items in JEI mean clutter, and if anything confusing for someone new wanting to use the mod here.
> > But being version independent from the mod is a huge factor to consider for sure.
> > I still think a clean JEI is preferred, but I can readd them if that's better as an add-on mod.
> 
> I think waiting for Int's opinion on this is the best move. There's pros and cons. It can also be quite easy to edit the file when Zorbatron DOES make some of it usable since I follow ZBGT closely and speak to Zorb daily. So it's not like ANY mod that we won't follow up on

I appreciate the initiative to hide unused items from JEI, but given the status of ZBGT as an addon mod and not a mod directly integrated into the pack, and its active development status, I think that these 'unused items' should probably stay in JEI.

However, I agree with D-Alessian that all items further than UHV should be removed. We can probably keep UHV items due to the existence of UHV converters and hatches (unless I am mistaken, and these do not exist).

overrides/groovy/postInit/addon/zbgt.groovy
Mazlaci: Removed a hiding for creative item bus.

overrides/groovy/postInit/addon/zbgt.groovy
Mazlaci: As per my other comment using metaitem for ZBGT stuff throws an error for me.
If it can be fixed I would be able to rewrite them.

overrides/groovy/postInit/addon/zbgt.groovy
Mazlaci: Simplified big parts of the code with loops.

overrides/groovy/postInit/addon/zbgt.groovy
Mazlaci: Readded general unused items, kept UEV+ stuff hidden.

IntegerLimit:
> I updated the code as of the requests.
> 
> I simplified it significantly using loop and variables. I don't know why I haven't thought of it originally.
> 
> ### I readded items to JEI:
> * Creative Item Bus
> * Misc Blocks, Wraps with no uses
> 
> ### Still hidden in JEI:
> * UEV+ stuff because of progression
> 
> ### Cells
> I kept the unused cells hidden, and also kept my simplification of the 60K Cell. I felt these fit in a diffenert category, and are better organised this way, but I can also readd them if needed,
> 
> ### Metaitem vs Item
> Originally I wanted to use metaitem and used it wherever I could still. But for some reason for every ZBGT item metaitem tags just dont work for me. Thats why I used the item solution, metaitem gives me the following error for example: image
> 
> I would gladly rewrite it to be cleaner with metaitem, but I'm not sure why it doesn't work, or how to fix it.

Because GT Copy doesn't work properly at this time, metaitems are slightly broken from other mods.

You have to add `zbgt:` in front of it, e.g. `metaitem('zbgt:wrapped.smd.capacitor')`.

Mazlaci:
> Because GT Copy doesn't work properly at this time, metaitems are slightly broken from other mods.
> 
> You have to add `zbgt:` in front of it, e.g. `metaitem('zbgt:wrapped.smd.capacitor')`.

That was indeed the problem.

I have rewritten it to use metaitem tags everywhere applicable.

IntegerLimit (approved):
LGTM, and thank you for the PR and making the changes requested.

@D-Alessian can you do a re-review based on the balancing aspects, since I am not familiar with ZBGT?

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -221,3 +221,171 @@ if (LabsModeHelper.normal) {
...
 }
+
+
+/* YOTTank Cells */
+// Add new recipes for YOTTank Cells
+mods.gregtech.assembler.recipeBuilder()
+	.inputs(
+		item('gregtech:hermetic_casing', 0),
+		metaitem('field.generator.lv'),
+		ore('circuitLv') * 4,
```
D-Alessian: LV feels a bit weird since YOTTank is EV? Also I don't see why you'd ever craft the LV one at EV. And it's weird to skip from HV to UHV. At the same time, T1 being 1 000 000L per cell would make it awful at EV. Rn YOTTank would have max (with 135 cells), at each tier (B = Buckets), using NM:
LV : 135 000 000 L / 135kB / ~16k cell (HV) / ~Quantum tank I (LuV)
HV: 13 500 000 000 / 13.5MB / ~ 1024k cell (IV) / ~ 6 Quantum tank V (UHV)
UHV: 1 350 000 000 000/ 1.35GB / ~ 10 16384k cells (ZPM) / ~600 Quantum Tanks V (UHV)

Which, imo, makes it pretty weak. At EV, making the full multi with LV cells is barely better than something that should be free for you atp, if you chose the HV cells, it's still only four times as good as an EV cell, but at least it's something. And at UHV, you'd just craft 10 16384ks instead of using up 270 field generators.

A potential solution would be to start at T2 and change the lang files so that T2 is EV, offering a good AE2 alternative, T3 is LuV where it would be really solid, probably cheaper than cells and offering a comparable storage amount, and T4 is UHV, where it would really be the ultimate storage solution for post tank, offering the equivalent of up to around 1000 1684k cells. It could be more expensive than that though, probably with an assline recipe. But the T2 would appear as T1, T3 would appear as T2, etc.

IntegerLimit:
No problems with the new modifications code-styling wise, up to @D-Alessian for his opinion on the balancing changes.

Mazlaci:
> There is also a failed check; to fix that, please fix the spelling of `probably` in `nomiceu.tooltip.yottank_cell.yottank_cell.2` lang, and update your branch with the latest commit in Nomi-CEu.

Fixed the spelling error, as far as I can see no more failed checks

IntegerLimit:
Thank you! Once @D-Alessian re-reviews, the PR is good to go!

Mazlaci:
As requested in my other PR (#1352), I changed the crafting matrix to not use any spaces.
It should be now up to standard.

IntegerLimit:
Thank you.

D-Alessian (approved):
I have not tested. But I assume it has been tested. The only thing I can't know from the code itself is if the langs are properly formatted, but that should be pretty obvious if this indeed has been tested.

Issue #1343, opened 2025-05-30
Fluid Extended Crafting Terminal skips progression steps


Mazlaci:
### Nomi CEu Version

1.7.4

### Launcher

CurseForge

### Configurations or Mods Changed

Added AE2 Fluid Crafting Rework

### Environment

Singleplayer

### Mode

Normal Mode

### Related Problem

Fluid Crafting mod kind of let's you cheat the extended AE2 terminals.
By not requiring items related to PackagedAuto, it's possible to do with your first EV Circuit.

Comparison for the "normal" and fluid extended terminal crafts:
![Image](https://github.com/user-attachments/assets/c4f48b44-c078-4743-82a5-ea969a461d4e)
![Image](https://github.com/user-attachments/assets/a7c4b124-a6f0-4142-bbc7-6987d78d8183)

Expecially big advantage is the ME Ultimate Encoder, allowing even bigger pattern essentially immediately after the first one:
![Image](https://github.com/user-attachments/assets/fc453e4c-e68b-4c3b-8515-33d277683b4d)

### Your Solution

Because Fluid Crafting is an official addon I think these blocks should be part of the proper progression.

### How Will Your Solution Work

I propose the Extended Fluid terminal should require a Package Recipe Encoder too instead of a Fluid Pattern Encoder.

Ultimate Encoder should be also become available later on.
It's a less clear cut answer when exactly, but I think in late game, when you unlock PackageExCrafting blocks feels appropriate.

IntegerLimit:
I agree with Mazlaci, the fluid extended should require package auto, just like the normal extended.

PR #1352, opened 2025-06-05
Update  AE2 Fluid Crafting Addon Script

Mazlaci:
Expanded version of the AE2 Fluid Crafting addon script with balancing changes.

### Zenscript to Groovy
Converted the original zenscript file to groovy script.
I followed the structure of the original file, and referenced the original author since the old file is deleted.

### Recipe changes
Containt the following recipe changes and additions.

#### Extended Fluid Pattern Terminal
As per my earlier issue #1343 I updated the crafting revice to use Packaged Recipe Encoder instead of the Fluid Recipe Encoder, bringing it in line with the "normal" Extended Pattern Teminal
image

#### Ultimate Encoder
Originally ths file was unchanged using the default craft. This meant you could immediatly upgrade to an even bigger pattern encoder, before even meddling into PackagedAuto.
My design goal was to add it to the proper progression, giving you the option to use it later on, approximately around the time you unlock PackagedExCrafting.
The new recipe uses the Ultimate Catalyst. It means yue require Uranium to craft it. It's not exact but around the intended time, late midgame, early lategame when you can also make the Ultimate Extended Crafting Table.
Not only was this designwise what I intended, but it's thematic for the Ultimate Encoder to use the Ultimate Catalyst.
image

#### Fluid Level Maintainer
With both Lazy AE2 and AE2 Extended Fluid Crafting downloaded, you have access to the Level Maintainer and the Fluid Level Maintainer. They both serve the same purpose for fluids and items, but with the original recipe it meant Fluid Level Maintainer was avalable much earlier at HV while Item Level Maintainer required complex comonents around IV
I felt it gives for a better progression, if you unlock these later on not effetively immediatly after AE2.
I made the new Fluid Level Maintainer recipe based on the Item Maintainer, without using blocks from an another addon of course. It now needs az IV circuite and more complex alloys, meaning the similar finction is reflected in a similar crafting recipe.
image
For reference Item Level Maintainer recipe:
image

#### Fluid Assembler
This one is more of a thematic addition then anything. This recipe was also unchanged originally, but the default recipe needed iron ingots and green dye, which just felt very "un-gregtech-like". It now requires iron plates and 1k ME Fluid Components, meaning it's just a slight nerf for a more integrated recipe in the progression.
image

### Metaitem vs Item
In the case of applied energetics 2, for the life of me I cannot get metaitem to work...
Original file used item tags, so I used them here too, but added comment for every one of them for easy identification.


overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
```diff
@@ -0,0 +1,112 @@
+// MODS_LOADED: ae2fc
+
+// Original Compatibility Author: tracer4b in file AE2FluidCrafting.zs
```
IntegerLimit: We can remove this (since this is all one project and credits are only provided to those outside of the org).

overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
```diff
@@ -0,0 +1,112 @@
+// MODS_LOADED: ae2fc
```
IntegerLimit, line 1: `// ^, Makes the script not run as long as AE2 FC isn't a mod in the player's instance.`

Please add this immediately beneath the `MODS_LOADED` comment, to clarify its purpose.

overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
```diff
@@ -0,0 +1,112 @@
...
+crafting.remove('ae2fc:fluid_level_maintainer')
+crafting.remove('ae2fc:part_fluid_pattern_ex_terminal')
+crafting.remove('ae2fc:ultimate_encoder')
+crafting.remove('ae2fc:fluid_assembler')
+
+// Fluid Discretizer recipe
+crafting.shapedBuilder().name('ae2fc:fluid_discretizer')
+	.output(item('ae2fc:fluid_discretizer'))
+	.matrix('EIE'
+		   ,'PGP'
```
IntegerLimit: The third level of indentation seems to be three spaces, it should still be a tab.

IntegerLimit (changes requested):
Thank you for the PR! Always nice to see scripts converted to Groovy, and you have done this very well.

See comments for some small formatting changes required.

Also, since I don't have easy access to LazyAE2 instance, could you quickly state the crafting recipe for the item level maintainer? Thanks in advance.

overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
Mazlaci: I added the requested explanation comment.

overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
Mazlaci: Removed the unneeded credit

overrides/groovy/postInit/addon/AE2FluidCrafting.groovy
Mazlaci: The reason I used spaces instead of a tab is to have the matrix rows under each other in an order.
I messed up while uploading the new file, using tab ident 8 instead of 4. Now if fix it, is it's as it should.

I can technically remove the space, but I would prefer to kepp it so it's visible at first glance.

IntegerLimit:
I would prefer if you used spaces as well for the last level of indentation; just to remain consistent with other files and our formatting specification (in `.editorconfig`).

Thanks in advance!

IntegerLimit (changes requested):
See above on relating to craft matrix formatting.

Mazlaci:
> See comments for some small formatting changes required.

I added the requested comment changes, and fixed the tab identation.
But as per my other comment the reason for 3 spaces is so the recipe matirx is aligned.

> Also, since I don't have easy access to LazyAE2 instance, could you quickly state the crafting recipe for the item level maintainer? Thanks in advance.

image

The recipe needs Fluix Steel Ingots, and Fluix Logic Unit
Tne Ingot is made from End Steel in the Alloy Smelter:r
image

The Logic Processor is a Circuit Assembler recipe with IV processor and titanium frame:
image

IntegerLimit (approved):
LGTM. Thank you for correcting all the comments so quickly!

PR #1371, opened 2025-06-18
Nerf Outputs of ME Conduit Recipes

### LINKED ISSUE #1321: changes regarding ME conduits
### Nomi CEu Version

Nightly

### Launcher

-

_No response_

### Environment

-

### Mode

Normal Mode

### Related Problem

me conduits cause bugs such as randomly disconnecting from blocks and me cables alike. this can cause a lot of confusion for players.

### Your Solution

make conduits only viable for setups that require the saved space from sharing a block with other conduits, f.e. by increasing their price.

currently you get 8 conduits per 3 cables, some rubber and some free cobbleworks resources in the form of conduit binder. make the recipe land you only 3 conduits per 3 cables. this would make the conduits just as affordable for the specific situations where they might be required but their cheapness wont incentivize their excessive use that leads to the aforementioned problems.

*taking into account only the assembler recipe since all players already have it by the time they start AE2

i'd considered suggesting the complete removal of ME conduits but got reminded that they still could be vital for space constrained setups such as small cleanrooms.

### Additional Information

_No response_

IntegerLimit:
This PR reduces the amount of ME Conduits created from the recipes:
1 from crafting (taking 3 covered cables, down from 4)
2 from assembler (taking 3 covered cables, down from 8)

This incentivises the use of Glass Cables, and deincentivizes the use of ME Conduits.

Fixes #1321

### DIFF
--- a/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
+++ b/overrides/config-overrides/expert/betterquesting/DefaultQuests.json
@@ -4451,7 +4451,7 @@
       "properties:10": {
         "betterquesting:10": {
           "autoclaim:1": 0,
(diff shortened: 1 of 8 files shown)


D-Alessian (approved):
A 4x nerf is maybe a bit much? But ig it is indeed the only way to make cables the cheaper option. Which is a good thing because using an option that ticks, can disconnect and has very taxing rendering should not be incentivized over the option that has none of these problems

PR #1376, opened 2025-06-27
Buff Cell Recipes

### LINKED ISSUE #1333: Drums and equivilent cells need balancing
### Nomi CEu Version

1.7.5

### Launcher

Prism Launcher

_No response_

### Environment

Singleplayer

### Mode

Both Modes

### Related Problem

_No response_

### Your Solution

The cell crafting recipe should stay the same, but it should craft 4 cells instead of 1

This would give players the choice of having placeable drums for a slightly higher price or craft 4 cells and not get in-world fluid storage

_No response_

### Additional Information

_No response_

IntegerLimit:
This PR increases the output amount of all cell recipes from 1 to 4, as well as changing the recycling recipes.

Fixes #1333

### DIFF
--- a/overrides/groovy/postInit/main/general/earlyGame/earlygame.groovy
+++ b/overrides/groovy/postInit/main/general/earlyGame/earlygame.groovy
@@ -1,6 +1,7 @@
 package postInit.main.general.earlyGame
 
 import com.cleanroommc.groovyscript.helper.ingredient.OreDictIngredient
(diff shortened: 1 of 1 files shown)


D-Alessian (approved):
I have no tested it, but I very much assume you have? 
As for the changes themselves I love it, makes cells more useful compared to other options that ignore their restrictions like super tanks or thermal's tanks. Cells being expensive doesn't really add anything interesting.

PR #1395, opened 2025-07-07
Improve ZBGT Integration; Cleanup ZBGT + AE2FC Scripts

IntegerLimit:
This PR makes several changes:

## Cleanup
- `AE2FluidCrafting.groovy` renamed to `ae2fc.groovy` to conform to repo standards
- ZBGT:
  - Use `changeBy` for field gen fix
  - Improve newline locations
  - Replace some repetitive code with loops
  - Algorithmic process for unwrap crafts
- Ports circuit recipe changes to GrS

## Bug Fixes
- Circuit unwrap recipes giving 16 circuit outputs; they take 1 circuit input (e.g. circuit duplication)
- Changed CoAL field gen recipe applying for Normal Mode
- Fix ULV Cover Recipes in CoAL
- Fix Circuit Assembly Line recipes for three circuits

## Testing To-Do
- Dedicated server testing, to ensure the property server-blocklist works correctly (with and without zbgt installed), as well as checking for other methods missing client side only annotations


overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -239,49 +236,39 @@ crafting.shapedBuilder().name('zbgt:yottank_cell_0')
...
-	.duration(100).EUt(VA[LuV])
-	.buildAndRegister()
+for (var tinLike : [fluid('tin') * (L * 2), fluid('soldering_alloy') * L]) {
+	mods.gregtech.assembler.recipeBuilder()
+		.inputs(
+			item('gregtech:hermetic_casing', 3),
+			metaitem('field.generator.hv') * 2,
+			ore('circuitEv') * 4,
+			metaitem('plateTitanium') * 4)
+		.fluidInputs(tinLike)
```
D-Alessian: this recipe didn't have tin/solder before. Not an issue, just thought I'd mention it in case this wasn't intentional

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -300,75 +287,42 @@ addTooltip(item('zbgt:yottank_cell', 0), [
 	translatable('nomiceu.tooltip.yottank_cell.yottank_cell.2')
 ])
 
+for (int i = 1; i <= 3; i++) {
```
D-Alessian: If this is made proceduraly, maybe adding a config option for up to what tier cells are generated COULD be interesting? Idk if it's possible via gs itself or if this would require a Labs update? Just an idea.

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -390,89 +344,151 @@ mods.jei.ingredient.removeAndHide(metaitem('zbgt:coolant_cell.60k.nak'))
...
+		ore('circuitUlv') * 48,
+		ore('stickLongBronze') * 24,
+		metaitem('ulv_covers:electric.piston.ulv') * 48,
+		metaitem('ulv_covers:electric.motor.ulv') * 48)
+	.circuitMeta(4)
+	.outputs(metaitem('ulv_covers:robot.arm.ulv') * 64)
+	.duration(600).EUt(VA[ULV])
+	.replace().buildAndRegister()
+
+mods.gregtech.coal_recipes.changeByOutput([metaitem('ulv_covers:conveyor.module.ulv')], null)
```
D-Alessian: I'd say, here, and in some other parts of the code, there's too much nesting it makes it pretty hard to understand what's actually going on. It might be somewhere where using changeByOutput might not be the most maintainable in my opinion and explicitly making each change might be better?

Just removeByOutput and recipeBuilder with a for loop for the rubber, seems just shorter and easier to read for something that ultimately just changes 3 recipes.

overrides/groovy/postInit/addon/zbgt.groovy
```diff
@@ -390,89 +344,151 @@ mods.jei.ingredient.removeAndHide(metaitem('zbgt:coolant_cell.60k.nak'))
...
+					it.copyWithAmount((int) (it.getAmount() / 2)) // Reduce motor amt to half
+				}.builder { RecipeBuilder recipe ->
+					recipe.clearFluidInputs()
+						.fluidInputs(rubber * (L * 2 * 48))
+						.circuitMeta(5)
+				}.buildAndRegister()
+		}
+	}
+
+// Field Generators (ZPM, UV (HM Only))
```
D-Alessian: this says UV (HM only) but neither of those recipes change a UV field gen

D-Alessian (commented):
Overall, I see no real big issue, except for the UV field gen which may be an oversight?

But, I feel like some of the code tries to be cleaner by doing everything with only one function call instead of multiple, and sometimes it's much cleaner, but sometimes it also makes it less maintainable in my, personal, opinion. 


Dedicated Server was tested with ZBGT and AE2FCr, works fine.

overrides/groovy/postInit/addon/zbgt.groovy
IntegerLimit: Yes, I feel that having tin/solder here makes sense since the higher tier recipe has it.

overrides/groovy/postInit/addon/zbgt.groovy
IntegerLimit: Not possible within GrS itself, and not currently done throughout the rest of nomi. Good idea though, perhaps we could consider more flexible configurable scripts later on.

overrides/groovy/postInit/addon/zbgt.groovy
IntegerLimit: Thank you for the comment. The reason I do prefer `changeBy...` methods is that they are more *declarative*, hence we are aware what is actually being changed in the recipe and so maintenance in terms of balancing is easier.

However, you are right that here, I probably overreached with the `changeBy` methods and a simple replacement would be better, considering all the other ULV recipes are simple replacements as well, and the aspect of adding different rubbers.

overrides/groovy/postInit/addon/zbgt.groovy
IntegerLimit: Thank you, will fix this later.

IntegerLimit: Fixed in latest.

overrides/groovy/postInit/addon/zbgt.groovy
IntegerLimit: Fixed in latest.

D-Alessian (approved):
LGTM. Would be nice to split it into multiple files, 500 loc is starting to get a bit much lol for a file like this with simple function calls and no deep logic, but if it should be done now or later, I think you'd know better.

Issue #1407, opened 2025-08-11
Liquid concrete cell not craftable


CrossElectronics:
### Nomi CEu Version

Nightly

### Launcher

Prism Launcher

### Configurations or Mods Changed

GregTech: 
```
I:additionalVeinsInSection=0
B:generateVeinsInCenterOfChunk=true
I:minVeinsInSection=1
```

### Environment

Singleplayer, nightly 3b71fd7, Java 8

### Mode

Expert Mode

### What Happened

<img width="1021" height="422" alt="Image" src="https://github.com/user-attachments/assets/6a1fe501-27d7-41b3-808c-5e123f88bf53" />
Concrete recipe always yields empty cells without any NBT tag. Liquid concrete cell has no valid recipe.

### Expected Behavior

Concrete recipe yields Liquid concrete cell.

### Reproduction Steps

1. Prepare a stack of empty cells and other materials.
2. Put in a crafting station and shift-click to craft 54 concrete cells
3. Empty cells are produced and can stack with the remaining 10 empty cells

### Logs

[crafttweaker.log](https://github.com/user-attachments/files/21712106/crafttweaker.log)

PR #1405, opened 2025-08-03
Manifest and Groovy Script Linting

IntegerLimit:
This PR adds linting capabilities to the manifest and groovy scripts. This is detailed below.

## Manifest
- Checks added to ensure sorted status
- Formatting is checked

## Groovy
- Configuration and integration of [npm-groovy-lint](https://www.npmjs.com/package/npm-groovy-lint)
- All existing problems resolved
- Files moved to folders with lowercase names, to adhere to convention of lowercase package names

All changes made to groovy files need to be checked to ensure it is correct. The scripts have been tested for errors, and some areas deemed volatile to breakage were tested in detail, but more testing is required by the reviewers.

**All other changes to groovy scripts is suspended until this PR is merged to avoid conflicts.**

PR #1408, opened 2025-08-11
Fix Concrete Cell Recipe

IntegerLimit:
This PR fixes the concrete cell recipe outputting a normal cell instead of a filled concrete cell. Also renames the fluid container ingredient to improve clarity.

This was due to https://github.com/Nomi-CEu/Nomi-CEu/commit/fe316b300d2b1b9996c970b252622a5ca2556f52, which modified the behaviour of `getHandler`, causing it to always copy stacks.

This resulted in the stack being returned by `fillStack` to be the original stack, with no modifications.

Fixes https://github.com/Nomi-CEu/Nomi-CEu/issues/1407


IntegerLimit:
This will cause a conflict with #1405, but this issue is critical enough to need this.

PR #1410, opened 2025-08-16
New Network Visualization Tool Recipe

D-Alessian:
The network visualization tool is currently EV gated, it's not really useful to gate it this far, as the only thing it does is help people who did turn channels on visualize their channels. I put it at MV, adding an alternate recipe using LV sensors.

I put this in the AE2 directory, since AE2Stuff adds a grand total of 4 items. This is opinionated so let me know if you think it should go elsewhere.

Old vs. New Recipe. Keeps both to not break patterns.
image

### DIFF
--- a/overrides/groovy/post/main/mod/ae2/items.groovy
+++ b/overrides/groovy/post/main/mod/ae2/items.groovy
@@ -39,6 +39,20 @@ crafting.shapelessBuilder()
     .input(ore('itemIlluminatedPanel'), item('actuallyadditions:item_laser_wrench'))
     .replace().register()
 
+// Network Visualisation Tool
+crafting.shapedBuilder()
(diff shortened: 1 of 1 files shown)


overrides/groovy/post/main/mod/ae2/items.groovy
```diff
@@ -39,6 +39,20 @@ crafting.shapelessBuilder()
...
     .replace().register()
 
+// Network Visualisation Tool
+crafting.shapedBuilder()
+    .output(item('ae2stuff:visualiser'))
+    .matrix(
+        'S S',
+        'EPE',
+        'FFF'
+    )
```
IntegerLimit: Following nomi conventions, this bracket should follow the string on the previous line.
```suggestion
        'FFF')
```

PR #1421, opened 2025-10-18
Misc QB fixes

v3ect0rgames:
* Sodium Persulfate & Iron III Quest shouldn't be locked behind Sulfuric Acid & Chlorine, should be either 
* Make it clear GP generation needs Atomic Reconstruct
* Quest for Clipboard
* You can wallshare DT, you just can't wallshare DT Hatches
* Mention EV Output Req for T2 MM
* Solidifier Quest shares Icon with Extractor Quest
* Mention that maintenance hatch in the middle of an edge for cleanroom can cause issues with forming


IntegerLimit:
Last point not relevant, has been fixed. Please fix the qb formatting issues (see the failed check)

PR #1457, opened 2025-12-16
[Addon] Deprecate ZBGT Generic Circuits

IntegerLimit:
This PR deprecates ZBGT generic circuits:
- Keeps their oredict and their assembler recipe (but hides the assembler recipe), preventing existing patterns from being broken
- Changes circuit unwrap recipe to give labs' universal circuit (only recipe that produces generics apart from its assembler recipe)
- Adds a deprecation tooltip
- **No direct conversion in crafting table**: not really needed given the zbgt generics can still be used, and would break things like generic LV -> programmed circ.

### DIFF
--- a/overrides/groovy/post/addon/zbgt/content.groovy
+++ b/overrides/groovy/post/addon/zbgt/content.groovy
@@ -3,18 +3,16 @@
 
 package post.addon.zbgt
 
-import com.nomiceu.nomilabs.groovy.ChangeRecipeBuilder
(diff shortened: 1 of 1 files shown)


IntegerLimit:
Good catch, but since circ number 26 is unused I believe; it should only very rarely be accidentally triggered (esp combined with how little circuits are used in assembler)

D-Alessian:
Oh absolutely, I'm just nitpicking for the sake of the review

PR #1464, opened 2025-12-22
HM: Remove Ench Books from Airtight Seal Recipes

IntegerLimit:
This PR removes the enchantment books respiration and holding from the HM recipe for the industrial rebreather kit, instead replacing them with a MV buffer and a MV super tank respectively.

This has a primary and a secondary benefit:
Primary: the rebreather kit recipe, and thus the airtight seal recipe, can be properly handled by ae2, and won't accept any enchanted book
Secondary: exposes players to the GT buffers

Theming:
- Buffer: intended as a 'temporary storage' of oxygen below the gas collector, perhaps close to one's oxygen inflow, thus similar to the respiration enchantment
- Tank: acts as an additional storage for oxygen, similar to the holding enchantment

Balancing:
- Buffer and tank are MV to prevent excess titanium req, and to match the voltage of the gas collector and chemical reactor

### DIFF
--- a/overrides/groovy/post/main/mod/advancedRocketry.groovy
+++ b/overrides/groovy/post/main/mod/advancedRocketry.groovy
@@ -56,12 +56,6 @@ if (LabsModeHelper.normal) {
         .duration(500).EUt(VA[HV])
         .buildAndRegister()
 } else {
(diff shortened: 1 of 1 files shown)


IntegerLimit:
Approved by @doniascion in discord.

PR #1465, opened 2025-12-22
Various AE2 Balancing Changes

IntegerLimit:
This PR makes various AE2, NAE2 and AE2 FC balancing changes.

### Hard Mode

**Early Progression:**
- Charger: Dark Steel -> Black Steel
- Crystal Growth: Blue Steel -> Black Steel

This makes it so that AE2 is locked behind black instead of blue steel; moving it a bit earlier: before the MV energy hatch (as blue steel needs HV in EBF) and before the MV mixer (as blue steel needs rose gold, which needs MV mixer)

It also means the fact that AE2 is locked behind black steel is easier to see, as its present in the very first block required (charger).

The legacy recipes have not been kept; players should be forced to migrate as they are completely different, and the new recipes are not inherently cheaper.

**Fluid Storage:**
- 1k Fluid Storage Component: Voltage: EV -> HV; Circuit Tier: EV -> HV
- 4k-64k Fluid Storage Component: Voltage: EV -> HV

This does have a slight side effect that the 64k component now doesn't require EV circuits, but given the amount of HV circuits required and the cost of the exquisite lapis, it is negligible.

The legacy recipes have not been kept; players should be forced to migrate as the savings from the new recipes will be worth the slight annoyance in migration.

**AE2FC:**
- Fluid Discretizer: Energetic Alloy Plates -> Titanium Plates; LV Field Gen -> HV Field Gen
- Fluid Pattern Encoder: MV Sensor -> EV Sensor
- Dual Interface: HV Circuit -> EV Circuit
- Fluid Assembler: Iron Plates -> Titanium Plates

This moves AE2FC to EV. Previously it was locked behind EV regardless due to the fluid components, but these changes properly lock it behind EV.

Note that we are using HV field gen instead of EV to prevent locking AE2FC after uranium processing. Thanks to @doniascion for this catch.

Legacy recipes have not been kept; players should be forced to migrate as the new recipes are more expensive.

### Normal Mode
**AE2FC:**
- Fluid Discretizer: Energetic Alloy Plates -> Stainless Steel Plates; LV Field Gen -> HV Field Gen
- Fluid Pattern Encoder: MV Sensor -> HV Sensor
- Fluid Assembler: Iron Plates -> Stainless Steel Plates

This moves AE2FC to HV; ensuring people have some AE2 system set up before they can dive into AE2FC, adding some progression to it and hopefully preventing extreme AE2FC abuse. Furthermore, fluids being required for autocrafting really only starts at HV.

Legacy recipes have not been kept; players should be forced to migrate as the new recipes are more expensive.

### Both Modes
**NAE2:**
- Beam Former: much cheaper now, just a crafting recipe (8 fluix wires + fluix pearl)
image

This is due to the beam former being simply just a replacement of a couple of wires and having a large limitation with needing line of sight; so it should not be expensive.

Suggested by @NotMyWing. The legacy recipes have not been kept; players should be forced to migrate as the savings from the new recipes will be worth the slight annoyance in migration.

**AE2 Stuff Wireless:**
- Wireless Connector: Iron Ingot -> Stainless Steel Plates; AE2 Wireless Receiver -> GT Wireless Transmitter; 2 Fluix Crystals -> 2 Beam Formers; Output Amount: 1 -> 2

Screenshot 2025-12-23 at 10 07 53 am

This moves the wireless connector a bit earlier; to early HV instead of mid EV. This places it ahead of progression from the beam former (which needs LOS and has a range of only 32) and behind of progression from the quantum ring (which requires more setup; but consumes much less energy over long distances and can cross dimensions).

The legacy recipes have not been kept; players should be forced to migrate as the new recipes may not be cheaper under all circumstances.

- Wireless Setup Kit: AE2 Wireless Receiver -> GT Wireless Transmitter

Part of the change moving wireless connector earlier.

Legacy recipe has been kept; as this players will likely only make this very sparingly, so we shouldn't break patterns.

- Advanced Wireless Setup Kit: Singularity -> 4k Storage Component

Part of the change moving wireless connector earlier. In theming, the storage component represents the advanced kit's ability to store multiple binding points.

The legacy recipes have not been kept; this item and its recipes only exist in nightly.

## Bug Fixes
Fixes a regression where 16k to 64k fluid components were crafting recipes instead of assembler recipes in HM.

## Other
- Adds a tooltip to AE2FC's Fluid Discretizer in RED: `Required for AE2 Fluid Crafting to work!`

### DIFF
--- a/overrides/groovy/post/addon/ae2fc.groovy
+++ b/overrides/groovy/post/addon/ae2fc.groovy
@@ -30,7 +30,7 @@ crafting.with {
         .key('E', LabsModeHelper.expert ? ore('plateTitanium') : ore('plateStainlessSteel'))
         .key('I', item('appliedenergistics2:part', 220)) // Storage Bus
         .key('P', item('appliedenergistics2:material', 24)) // Engineering Processor
-        .key('G', LabsModeHelper.expert ? metaitem('field.generator.ev') : metaitem('field.generator.hv'))
+        .key('G', metaitem('field.generator.hv'))
         .key('F', item('appliedenergistics2:part', 221)) // Fluid Storage Bus
         .register()
 


D-Alessian (approved):
LGTM. 

I do see a lot of ternary operators with expert mode, and I wonder if it could use some helper? Since the logic is really just always "if expert, use material of this tier, if not, use the same kind of material, but a tier up". It could be helpful to clarify which materials are "equal" to one another, but different by a tier, like dark steel and black steel. This could make it easy to change all of them at once.

This is a nitpick and an idea for 1.8 though. Hence why I approve.

PR #1471, opened 2025-12-23
Remove Prerequisite For Tiny Fuels Quest

v3ect0rgames:
Removes the prerequisite for tiny fuels quest.


IntegerLimit:
I have a couple more changes planned relating to this; not exactly removing prereq as done here. I'll do it in `qb-misc-1.7.6`.

PR #1473, opened 2025-12-23
Fix Multiblock Preview Quest

v3ect0rgames:
The Multiblock Preview Quest had issues about which feature did what.


IntegerLimit (approved):
We can tweak this in dev.

Issue #1491, opened 2026-01-01
[ADDON] Add recipes for Flux Networks' builtin energy storages


MasterKogha:
### Nomi CEu Version

1.7.6

### Launcher

N/A

### Environment

N/A

### Mode

Both Modes

### Related Problem

While FluxNets has an internal buffer per-i/o-block, it isn't amazing at knowing how much energy it needs to buffer and generally sometimes just cant keep up with multis sometimes (I'm not *totally* sure why, but sometimes flux points just decide not to output enough power unless you increase their priority or something, anyway giving the network a proper internal buffer via the built-in FluxNets storages fixes it also, but they're uncraftable here).

### Your Solution

Add recipes to the 3 storages (or at least 1 of them, it doesn't really matter you just need at least one of them available (though ideally at least the biggest one as energy usage here gets really quite big)).

### How Will Your Solution Work

As explained, as I've tested, adding one to the network to act as a real net-wide buffer completely addresses the issue (and is, IMO, good practice for all energy systems across all packs but that's besides the point). There's no balance issues at play here, the FluxNets storages are REALLY bad at actually storing energy (1/8/128MRF per storage block per tier, [mod as a whole is] LuV gated; compared to EIO's 1/5/25MRF per capacitor per tier, *LV* gated (give or take)). The mod loses to 6 vibrant capacitor banks at storing energy, let alone DE ball/PSS, this would have no effect on energy balance, it just fixes this transfer issue.

### Alternatives You've Considered

Modifying FluxNets' code sounds astronomically harder than this change.

### Additional Information

Also adjacent problem: a lot of FluxNets is disabled to be replaced with GT stuff, but isn't hidden from JEI. Specifically: Flux Dust (it's crafting "recipe" also could do with being disabled if easily possible), Flux Block, Flux Core, Admin Configurator (may want to be left in JEI similar to other creative debug tools), and the 3 storages that this issue is about (in case the change is rejected).

I'm happy to write and PR the scripting for this change myself, I just need to know what items I should set in the recipe(s) in question. Ping me `@masterkogha` on discord in the Nomi server (or reply here) if you want to discuss.

PR #1495, opened 2026-01-02
Various Flux Networks Adjustments

MasterKogha:
This PR applies some general tweaks to the Flux Networks integration (I wasn't adept enough at GS to convert the CT scripting at this stage unfortunately, but all feature changes are good to go).

Detailed changelog:
-
- Adds recipes to Flux Networks' built-in energy storage options, as adding one to a network helps it buffer energy and avoid odd behaviour inconsistent with it's stated "unlimited throughput"
- Hides flux, flux core, flux block, and their recipes from JEI as they are made redundant by the GT'ed recipes
- Flux point recipe flipped vertically to put the sensor on top, for consistency with all other recipes having the GT component(s) on top
- Flux plug sensor swapped for emitter, as it's the "sending" block, and the emitter is the "sending" GT tiered component (they cost the literal exact same resources)

Latter 2 changes break patterns but they're easily fixed (and very few players actually play with this mod specifically, as it's *really* strong), otherwise no issues 
Addresses #1491

### DIFF
--- a/overrides/groovy/post/addon/FluxNetworks.groovy
+++ b/overrides/groovy/post/addon/FluxNetworks.groovy
@@ -157,14 +157,11 @@ crafting.shapedBuilder()
 // Flux Storages
 // Specify that they help with "network stability"
 addTooltip(item('fluxnetworks:fluxstorage'), [
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.1'),
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.2'),
+    translatable('nomiceu.tooltip.addon.ae2fc.storages'),
 ])
 addTooltip(item('fluxnetworks:herculeanfluxstorage'), [
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.1'),
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.2'),
+    translatable('nomiceu.tooltip.addon.ae2fc.storages'),
 ])
 addTooltip(item('fluxnetworks:gargantuanfluxstorage'), [
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.1'),
-    translatable('nomiceu.tooltip.addon.ae2fc.storages.2'),
+    translatable('nomiceu.tooltip.addon.ae2fc.storages'),
 ])
--- a/overrides/resources/modpack/lang/en_us.lang
+++ b/overrides/resources/modpack/lang/en_us.lang
@@ -156,5 +156,4 @@ nomiceu.tooltip.xtones.lamp=§eRequires a redstone signal to light.§r
 nomiceu.tooltip.addon.ae2fc.discretizer=§cRequired for AE2 Fluid Crafting to work!§r
 
 # FluxNetworks
-nomiceu.tooltip.addon.ae2fc.storages.1=§eIt is recommended to add at least one Flux Networks Storage to each §r
-nomiceu.tooltip.addon.ae2fc.storages.2=§enetwork, as the buffer makes flux points push energy more reliably.§r
+nomiceu.tooltip.addon.ae2fc.storages=Add to your network for increased §dnetwork stability§r!


overrides/scripts/AddonScripts/FluxNetworks.zs
```diff
@@ -5,9 +5,7 @@ import scripts.common.makeShaped as makeShaped;
 print("--- loading FluxNetworks.zs ---");
 
 //Remove all default recipes
-recipes.remove(<fluxnetworks:fluxcore>);
 recipes.remove(<fluxnetworks:fluxconfigurator>);
-recipes.remove(<fluxnetworks:fluxblock>);
```
IntegerLimit, line 10: **Question**: what's the reasoning behind this?

overrides/groovy/post/addon/FluxNetworks.groovy
```diff
@@ -0,0 +1,10 @@
+// MODS_LOADED: fluxnetworks
+// ^, Only runs when a player adds Flux Networks themselves
+
+package post.addon
+
+// Remove category for flux dust in-world craft and hide all 3 no longer used items
+mods.jei.category.remove('flux')
+mods.jei.ingredient.hide(item('fluxnetworks:flux'))
+mods.jei.ingredient.removeAndHide(item('fluxnetworks:fluxcore')) // Have to removeAndHide as it shows up for uses/eye of ender/obsidian otherwise
```
IntegerLimit: Our CI is reporting this line (line 9) as too long. Perhaps move the comment to the line before.

overrides/scripts/AddonScripts/FluxNetworks.zs
```diff
@@ -53,6 +51,48 @@ makeShaped("of_flux_controller", <fluxnetworks:fluxcontroller>,
...
+     "XIX"],
+    { C : <ore:circuitLuv>, //T6 Circuit
+      S : <metaitem:sensor.luv>, //LuV Sensor
+	  E : <metaitem:emitter.luv>, //LuV Emitter
+      X : <enderio:item_endergy_conduit:11>, //Superconductor Wire
+	  I : <actuallyadditions:block_phantom_energyface>, //AA Phantom Energyface
+	  B : <metaitem:battery_buffer.luv.4>, //LuV Battery Buffer 4x
+	  A : <metaitem:energy.lapotronic_orb> //Lapotronic Energy Orb
+      });
+
```
IntegerLimit: **Balancing:** These recipes look nice, but perhaps we could communicate their increasing tier through more than a change in the energy orb? Perhaps it could include the storage of the previous tier (so players can also easily upgrade lower tier storages), and/or the tiers of the items used (e.g. LuV Sensor, Emitter) could increase in tier.

IntegerLimit (changes requested):
Thank you for the PR! The general implementation looks great.

Please see below for a couple comments.

Also, a couple general changes I would like to request:

## Change 1

> Latter 2 changes break patterns but they're easily fixed (and very few players actually play with this mod specifically, as it's really strong)

From what I can see, these two changes are purely for thematic purposes. We shouldn't break people's patterns, and we should instead keep the previous recipe, but mark it as 'deprecated'.

See [this groovy file](https://github.com/Nomi-CEu/Nomi-CEu/blob/main/overrides/groovy/post/main/mod/ae2stuff.groovy#L21-L37) for an example of how to do this. Note that this must be in groovy, so you might want to move the non-deprecated recipes over anyways (although that isn't required).

## Change 2

The information that the energy storages can result in more stable networks should be presented to players. Perhaps we can add a tooltip on the energy storages saying this?

See the [ae2fc groovy addon script](https://github.com/Nomi-CEu/Nomi-CEu/blob/main/overrides/groovy/post/addon/ae2fc.groovy#L124) for how to do this. Note you will need to add a line in the lang file [here](https://github.com/Nomi-CEu/Nomi-CEu/blob/main/overrides/resources/modpack/lang/en_us.lang#L156).

overrides/groovy/post/addon/FluxNetworks.groovy
```diff
@@ -0,0 +1,170 @@
...
+    .matrix(
+        'XSX',
+        'CFC',
+        'XUX')
+    .key('X', item('enderio:item_endergy_conduit:11')) // Drac. Superconductor Conduit
+    .key('S', metaitem('sensor.luv')) // LuV Sensor
+    .key('C', ore('circuitLuv')) // T6 Circut
+    .key('F', item('actuallyadditions:block_phantom_energyface')) // Phantom Energyface
+    .key('U', metaitem('energy_converter.luv.16')) // 16x LuV Converter
+    .register()
```
IntegerLimit, line 61: Not sure if this is intentional or not, but the recipe changed from 4x LuV circuits + 2x superconductor conduits to 2x LuV Circuits + 4x superconductor conduits.

A similar change occurred in the flux plug recipe.

overrides/groovy/post/addon/FluxNetworks.groovy
```diff
@@ -0,0 +1,170 @@
...
+    .output(item('fluxnetworks:fluxplug'))
+    .matrix(
+        'CSC',
+        'XFX',
+        'CUC')
+    .key('C', ore('circuitLuv')) // T6 Circut
+    .key('S', metaitem('sensor.luv')) // LuV Sensor
+    .key('X', item('enderio:item_endergy_conduit:11')) // Drac. Superconductor Conduit
+    .key('F', item('actuallyadditions:block_phantom_energyface')) // Phantom Energyface
+    .key('U', metaitem('energy_converter.luv.16')) // 16x LuV Converter
```
IntegerLimit: Can we maybe put these key declarations in the same order as the actual recipe so the difference is more obvious?

In this case: conduit, then sensor, then circuit, then phantom energyface, then converter.

Same for the legacy flux point recipe.

overrides/groovy/post/addon/FluxNetworks.groovy
```diff
@@ -0,0 +1,170 @@
...
+// Flux Controller
+crafting.remove('fluxnetworks:tile.fluxnetworks.fluxcontroller')
+crafting.shapedBuilder()
+    .output(item('fluxnetworks:fluxcontroller'))
+    .matrix(
+        'SPE',
+        'CBC',
+        'XFX')
+    .key('S', metaitem('sensor.luv')) // LuV Sensor
+    .key('P', item('extrautils2:playerchest'))
```
IntegerLimit: Missing item descriptor comment.

IntegerLimit (commented):
Thank you so much for the changes! The fixing of the nbt recipes is amazing, as is the full conversion from CT to GS.

Please see below for some minor formatting changes that do need to be applied.

Also, in your item descriptor comments, you've consistently wrote LuV circuit as `T6 Circut`, perhaps we can correct this to `LuV Circuit`?

overrides/groovy/post/addon/FluxNetworks.groovy
IntegerLimit: Oh I see, this was to prevent conflicts between the new flux point recipe and the deprecated legacy flux plug recipe.

MasterKogha: Yep (it also aligns with the rest of the recipes having conduit in corners, circuit in the middle, but mostly the conflict avoidance)

overrides/groovy/post/addon/FluxNetworks.groovy
MasterKogha: Those 2 recipes still have the old "circuit corner" setup, so this *is* in order for the legacy matrices

IntegerLimit:
Also, a note for the future: the item descriptor comments aren't really needed in *all* cases. When the function describes the item well (e.g. `metaitem`, or most `item` cases), its not needed. Its still preferable when its more vague (e.g. `item('appliedenergistics2:material', 15)` ), but still not required (mostly because https://github.com/IntegerLimit/GroovyScriptPlugin exists).

overrides/resources/modpack/lang/en_us.lang
IntegerLimit, line 1: Sorry, final request:

I don't like this tooltip line split, its not along a logical place, and I feel like the tooltip is a bit long.

Can we have this instead?

`Add to your network for increased §dnetwork stability§r!`

PR #1518, opened 2026-02-15
make easy mode a bit easier

FijiAura:
eu to fe (and vise versa) have been reduced slightly 
removed harder iron recipes because lost cities


D-Alessian:
You do this on many different GT related repos. Submit low quality (maybe troll) PRs and issues that clearly don't offer much of anything. 

No, we are not changing the eu to rf ratio. Changing flint and steel recipe is barely even relevant.

PR #1525, opened 2026-03-03
Sponge groovy compat

EffGod:
This PR updates the Groovy overrides so the same scripts load cleanly on both SpongeForge and Forge for 1.12.2.

What changed:
- replaced remaining obfuscated `func_*` calls in the Groovy overrides with readable helper-based logic
- added a safer `ingredientMatchingStacks(...)` implementation in `post/classes/Common.groovy` that resolves matching stacks without relying on unstable obfuscated `Ingredient` internals
- updated NBT handling in affected scripts to use shared helpers like `nbtHasNumericKey(...)`, `getTag(...)`, `copyTag(...)`, `setStackTag(...)`, and safer high-level paths like `withNbt(...)`
- cleaned up Groovy lint issues so the branch passes CI

Why:
- some previous fixes depended on runtime-obfuscated calls, which were harder to maintain and not acceptable for the Groovy checks
- the new helper-based approach keeps the code readable while still avoiding the runtime resolution problems that differ between SpongeForge and plain Forge

### DIFF
--- a/overrides/groovy/post/main/general/early/earlygame.groovy
+++ b/overrides/groovy/post/main/general/early/earlygame.groovy
@@ -362,13 +362,13 @@ crafting.shapedBuilder()
 // Basic Cell Mass Change
 mods.gregtech.bender.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * (out.count * 4) }
+        builder.changeEachOutput { out -> out * (out.amount * 4) }
             .replaceAndRegister()
     }
 
 mods.gregtech.extruder.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * (out.count * 4) }
+        builder.changeEachOutput { out -> out * (out.amount * 4) }
             .replaceAndRegister()
     }
 
--- a/overrides/groovy/post/main/general/early/earlygame.groovy
+++ b/overrides/groovy/post/main/general/early/earlygame.groovy
@@ -362,13 +362,13 @@ crafting.shapedBuilder()
 // Basic Cell Mass Change
 mods.gregtech.bender.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * ((out.hasProperty('amount') ? out.amount : out.count) * 4) }
+        builder.changeEachOutput { out -> out * ((out.hasProperty('count') ? out.count : out.amount) * 4) }
             .replaceAndRegister()
     }
 
 mods.gregtech.extruder.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * ((out.hasProperty('amount') ? out.amount : out.count) * 4) }
+        builder.changeEachOutput { out -> out * ((out.hasProperty('count') ? out.count : out.amount) * 4) }
             .replaceAndRegister()
     }
 
--- a/overrides/groovy/post/main/general/early/earlygame.groovy
+++ b/overrides/groovy/post/main/general/early/earlygame.groovy
@@ -362,13 +362,13 @@ crafting.shapedBuilder()
 // Basic Cell Mass Change
 mods.gregtech.bender.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * (out.amount * 4) }
+        builder.changeEachOutput { out -> out * ((out.hasProperty('amount') ? out.amount : out.count) * 4) }
             .replaceAndRegister()
     }
 
 mods.gregtech.extruder.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * (out.amount * 4) }
+        builder.changeEachOutput { out -> out * ((out.hasProperty('amount') ? out.amount : out.count) * 4) }
             .replaceAndRegister()
     }
 
--- a/overrides/groovy/post/main/general/early/earlygame.groovy
+++ b/overrides/groovy/post/main/general/early/earlygame.groovy
@@ -6,7 +6,7 @@ import static com.nomiceu.nomilabs.groovy.GroovyHelpers.TranslationHelpers.trans
 import static com.nomiceu.nomilabs.groovy.GroovyHelpers.GTRecipeHelpers.toGtInput
 import static gregtech.api.GTValues.*
 import static org.apache.commons.lang3.tuple.Pair.of
-import static post.classes.SpongeAdaptor.*
+import static post.classes.SpongeAdapter.*
 
 import com.cleanroommc.groovyscript.helper.ingredient.OreDictIngredient
 import com.nomiceu.nomilabs.groovy.ChangeRecipeBuilder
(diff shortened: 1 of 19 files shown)


v3ect0rgames (changes requested):
A few things to fix, make sure you aren't copying the readme from your repo into this one, try to avoid using the obfuscated functions, and remove double line spaces in ae2 groovy.

v3ect0rgames (approved):
This looks good as long as its tested to work on all versions. If the change from count to amount works on both forge and sponge, then it should be good.

EffGod:
The added implementation of count / amount variable preserves original usage of count. 
All changes made preserve original usage - added sponge specific fixes IF original properties fail.

v3ect0rgames (approved):
Alright looks good, but just make sure each mods functionallity, and all the recipes edited still work.

overrides/groovy/post/main/general/misc/qol/pulps.groovy
```diff
@@ -39,8 +38,8 @@ for (var material : materials) {
 
         mods.gregtech.extruder.changeByOutput([stack], null)
             .forEach { ChangeRecipeBuilder builder ->
-                builder.changeInput(0) { GTRecipeInput input ->
-                    new GTRecipeItemInput(dust, input.amount)
+                builder.changeInput(0) {
+                    new GTRecipeItemInput(dust, dust.hasProperty('amount') ? dust.amount : dust.count)
```
IntegerLimit: This still doesn't solve the issue. This copies the amount from the input dust; when it should copy the amount from the GTRecipeInput object.

This will still create incorrect recipes for everyone.

overrides/groovy/post/main/general/early/earlygame.groovy
```diff
@@ -362,13 +362,13 @@ crafting.shapedBuilder()
 // Basic Cell Mass Change
 mods.gregtech.bender.changeByOutput([metaitem('fluid_cell')], null)
     .forEach { ChangeRecipeBuilder builder ->
-        builder.changeEachOutput { out -> out * (out.count * 4) }
+        builder.changeEachOutput { out -> out * ((out.hasProperty('count') ? out.count : out.amount) * 4) }
```
IntegerLimit: Checking whether an ItemStack has property `count`, and using `count` or `amount` respectively seems to appear everywhere. A global util method could be useful here.

IntegerLimit (changes requested):
Thank you for your work in making this PR; it is much appreciated!

Generally, I'm still very concerned at how much we need to obfuscate the codebase, and the many extra considerations, that it takes to support sponge; especially considering we have to lose some features.

Perhaps it is best to wait until GroovyScript provides better integration with sponge before we properly officially provide support? However, if we can clean it up a bit more, perhaps this could work.

The first comment is my main concern; and I would like an explanation of what the two new util methods do, and why they are needed. The second comment is of lesser priority and should only be applied if we make a final decision to merge this PR.

Once again, thank you very much for your work.

overrides/groovy/post/main/general/misc/qol/pulps.groovy
IntegerLimit: I'm actually not sure why this should need changing at all; since that is not a MC object, but one from GT; so it shouldn't be changed by sponge's mixins right?

overrides/groovy/post/main/mod/ae2stuff.groovy
```diff
@@ -90,7 +90,7 @@ crafting.shapedBuilder()
 for (String type : ['ae2stuff:wireless', 'ae2stuff:wireless_hub']) {
     // Init chisel group
     mods.chisel.carving.addGroup(type)
-    mods.chisel.carving.setSound(type, SoundEvents.BLOCK_STONE_PLACE)
+    //mods.chisel.carving.setSound(type, SoundEvents.BLOCK_STONE_PLACE)
```
D-Alessian, line 93: It's a bit useless to comment out lines of code on git. Just remove it if it's not used anymore. 

I also feel like this straight up removes a feature? Maybe I'm wrong but I don't see where this gets reused.

D-Alessian (changes requested):
I'm only leaving a review based on the code itself, not testing nor knowledge about Sponge. I have not tested this, and I don't know much about Sponge either. Purely reviewing based on the shape of the code.

My overall feeling is that this code is bottom-up rather than what I think should be a top-down approach. Are we certain this covers all issues when using sponge? How are we going to proceed if more issues arise? And maybe I'm wrong and you did approach it with a bottom-up approach, the code just does not feel like it.

I also see a lot of code duplication, and this duplication isn't even always consistent which makes this very hard to maintain. 

The example Int raised with the count/amount properties is a good example. This appears 9 or 10 times in the code, and absolutely would be better with some method that takes in an ItemStack and returns an Int. I feel like this should be the main way forward: 
1. What breaks Sponge when using groovyscripts?
2. How can we fix it? Make methods to fix it in an adapter class.
3. Where is it used? Use the methods.

Also, for code clarity, I feel like it would be clearer if Sponge-specific helper methods were in a dedicated class like `post/classes/SpongeCompat.groovy` or `post/classes/SpongeAdapter.groovy` or similar.

overrides/groovy/post/main/mod/ae2stuff.groovy
EffGod: This was commented out because I dont want to 100% get rid of it.
and it is getting rid of the case of chisel sound on the wireless hub.
which when would someone ever chisel a wireless_hub anyway.

This wouldnt be touched if it didnt have to be.

overrides/groovy/post/main/general/misc/qol/handFraming.groovy
```diff
@@ -60,10 +60,9 @@ for (ItemStack stack : items) {
 }
 
 static ItemStack addNBT(ItemStack stack, boolean trim, boolean front) {
-    var sideStack = item('xtones:zane')
-    var trimStack = trim ? item('extendedcrafting:storage', 4) : ItemStack.EMPTY
-    var frontStack = front ? item('xtones:zane', 15) : ItemStack.EMPTY
-
```
D-Alessian, line 66: Why was this line break removed?

overrides/groovy/post/main/mod/thermal.groovy
```diff
@@ -114,7 +115,7 @@ class UpgradeIngredient extends SimpleIIngredient {
         if (!ItemMeta.compare(toUpgrade, itemStack)) return false
 
         // Check Level
-        var tag = itemStack.tagCompound
+        var tag = itemStack.hasProperty('tagCompound') ? itemStack.tagCompound : getTag(itemStack)
```
D-Alessian, line 118: same as for count, this could use a method in the adapted. Even if it's only used once for now, means less duplication later on. And would be useful if we try to get a type aware linter.

overrides/groovy/post/main/general/misc/qol/pulps.groovy
```diff
@@ -37,11 +34,8 @@ for (var material : materials) {
         var stack = OreDictUnifier.get(prefix, material)
         if (stack.empty) continue
 
-        mods.gregtech.extruder.changeByOutput([stack], null)
-            .forEach { ChangeRecipeBuilder builder ->
-                builder.changeInput(0) { GTRecipeInput input ->
-                    new GTRecipeItemInput(dust, input.amount)
-                }.buildAndRegister()
-            }
+        mods.gregtech.extruder.changeByOutput([stack], null).each { builder ->
```
D-Alessian, line 37: Why was this changed to a each loop instead of forEach?

D-Alessian (commented):
Overall it looks a lot better. I'd need to get my hands on the codebase a bit more to really give an opinion about some of these, but I probably will not be able to for 2 weeks at least.

I'll try and see about adding a type aware linter check to make sure these compats are always applied, so we don't break compat with some random updates.

PR #1532, opened 2026-03-21
Fix Load Order for Universal Circuit Changes 

### LINKED ISSUE #1526: Processing Array doesn't use universal circuits
### Nomi CEu Version

Nightly

### Launcher

Prism Launcher

Added AE2FCR 2.6.6-r
Added ZBGT 0.16.2

### Environment

Nightly, multiplayer server

### Mode

Expert Mode

### What Happened

Processing Array has LuV circuits that accepts any: circuitExclUniversalLuv
You can't use universal circuit

### Expected Behavior

To be able to use the universal circuits that created using forming press

### Reproduction Steps

Search for the Processing Array crafting recipe in JEI and move the cursor over the circuit with advanced tooltip on

### Logs

_No response_

### Additional Information

This is likely caused because the groovy script that contains this recipe change is in the file ".../groovy/post/main/general/mid/midgame.groovy" is loaded before the universalCircuits.groovy script.

The issue was solved if I loaded the recipe change after the universalCircuits.groovy

Image

D-Alessian:
Fixes #1526

The script to use universal circuits loaded after some of the scripts in `post/main`, as the issue said. This caused the processing array recipe (and maybe other recipes too) to use the `circuitExclUniversal<Tier>` oredic instead of `circuit<Tier>`. I created a new folder `post/boostrap` for scripts which need to load before the rest. 

Not entirely convinced about the name though, feel free to give feedback on that. I'm really bad at naming my folders.

### DIFF
--- a/overrides/groovy/post/bootstrap/general/misc/content/universalCircuits.groovy
+++ b/overrides/groovy/post/bootstrap/general/misc/content/universalCircuits.groovy
@@ -1,4 +1,4 @@
-package post.main.general.misc.content
+package post.bootstrap.general.misc.content
 
 import static gregtech.api.GTValues.*
 
--- a/overrides/groovy/runConfig.json
+++ b/overrides/groovy/runConfig.json
@@ -6,6 +6,7 @@
   "loaders": {
     "postInit": [
       "post/classes",
+      "post/bootstrap",
       "post/main",
       "post/addon"
     ]


IntegerLimit (approved):
Not too sure what the folder should be called either, but `bootstrap` works.

LGTM.

PR #1542, opened 2026-04-06
Pin 3rd Party Actions to Commit Hashes

D-Alessian:
This PR pins all 3rd party actions to a commit hash. Version tags are mutable and thus, prone to supply-chain attacks if the actions ever get compromised. This happened recently with [trivy-action](https://github.com/aquasecurity/trivy-action), causing a [massive supply-chain attack](https://snyk.io/blog/poisoned-security-scanner-backdooring-litellm/). This affects:
- crate-ci/typos
- Nomi-CEu/SecretChecker
- peter-evans/create-or-update-comment
- peter-evans/create-pull-request
- peter-evans/find-comment
- stefanzweifel/git-auto-commit-action

To keep the actions up to date without pinning to a version tag, I added a dependabot config in `.github/dependabot.yml`. It would be useful to take a look at it following the [config references](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference), and give feedback. Some consideration I had:
- use the `allow` option and `ignore` option with `update-types`? Might want to ignore breaking changes for example. I think this is the most important one to consider, but I left it open for now because it's pretty opinionated as to which strategy exactly should be used. How this would work in practice can be found [here](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated)
- use the `groups` option to avoid opening 100 PRs every week, perhaps? Though if it's just for github-actions it should be fine
- also update npm packages? It will keep the existing security updates behavior, this would be for regular non security related updates. Not sure if it's useful and could create a lot of PR bloat.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,12 @@ updates:
       - "github-actions"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "npm"
+    directory: "/tools"
+    labels:
+      - "dependencies"
+      - "auto-creation"
+      - "javascript"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,9 @@ updates:
       - "github-actions"
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
 
   - package-ecosystem: "npm"
     directory: "/tools"
@@ -15,5 +18,9 @@ updates:
       - "dependencies"
       - "auto-creation"
       - "javascript"
+    open-pull-requests-limit: 10
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    labels:
+      - "dependencies"
+      - "auto-creation"
+      - "github-actions"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,16 @@ updates:
       - "github-actions"
     schedule:
       interval: "weekly"
-    ignore:
-      - dependency-name: "*"
-        update-types: ["version-update:semver-major"]
+    groups:
+      actions-breaking:
+        applies-to: version-updates
+        update-types:
+          - major
+      actions-safe-updates:
+        applies-to: version-updates
+        update-types:
+          - minor
+          - patch
 
   - package-ecosystem: "npm"
     directory: "/tools"
@@ -18,9 +25,13 @@ updates:
       - "dependencies"
       - "auto-creation"
       - "javascript"
-    open-pull-requests-limit: 10
     schedule:
       interval: "weekly"
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
+    groups:
+      npm-safe-updates:
+        applies-to: version-updates
+        patterns:
+          - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,11 +27,14 @@ updates:
       - "javascript"
     schedule:
       interval: "weekly"
-    ignore:
-      - dependency-name: "*"
-        update-types: ["version-update:semver-major"]
     groups:
+      npm-breaking:
+        applies-to: version-updates
+        update-types:
+          - major
+
       npm-safe-updates:
         applies-to: version-updates
-        patterns:
-          - "*"
+        update-types:
+          - minor
+          - patch
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -63,12 +63,12 @@ jobs:
         run: npm run gulp buildTypoCheck
 
       - name: Lang Typo Check
-        uses: crate-ci/typos@v1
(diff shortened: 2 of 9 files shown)


IntegerLimit (changes requested):
Looks great.

Whilst we are at it though, maybe we can enable dependabot package updates for npm? I think the PR bloat will be minimal, and should be a net benefit. Grouping dependabot PRs together would also be great.

I think ignoring breaking changes is also a good idea; but I'm afraid it will keep us having our actions be very behind, so maybe we should just keep it as is. Most likely the updated actions will run on the dependabot PR and be tested there anyways.

D-Alessian:
> I think ignoring breaking changes is also a good idea; but I'm afraid it will keep us having our actions be very behind, so maybe we should just keep it as is. Most likely the updated actions will run on the dependabot PR and be tested there anyways.

I forgot to read the second part of this, I just read "good idea" lol. I'll remove them then for github-actions. We should also discuss the versioning strategy used. There are 5 options ([source](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#ignoring-specific-versions-or-ranges-of-versions)):

- `auto`: self explanatory. Decides on case per case.
- `lockfile-only`: doesn't change `package.json`, meaning the restrictions would not change. I feel like this is quite nice to avoid any breaking changes, and would mean we do not have to ignore semver changes, since that will already be defined by the `package.json`
- `increase`: always increases the restrictions in the lockfile. Seems risky imo. Especially without ignoring semver major.
- `increase-if-necessary`: same, but only if the previous restrictions do not allow for the update.
- `widen`: I don't really understand the purpose of this one honestly.

Also forgot to group them. Will do this right now.

Also think it would be good to test this in a fork before merging (no linter, and been a while since I've done this), so I'll do that too.

D-Alessian:
Tested dependabot up in https://github.com/Horde-Of-Greg/Nomi-CEu-HOG/pulls

Seems like using `lockfile-only` or ignoring updates to `typescript` would be necessary though, as updating typescript breaks the tools currently.

IntegerLimit:
Great. I'll see if I can fix the typescript updating error before we merge.

D-Alessian:
> Great. I'll see if I can fix the typescript updating error before we merge.

Oh it's not difficult, just two ways around it. Either use `lockfile-only`, which means it will never update the restrictions in `package.json` (for more than just typescript, but this might be desired), or just ignore the typescript updates altogether via `ignore:`. 

It's not particularly hard, just opinionated.

IntegerLimit:
I'll prefer to update it.

**UPDATE:** The issue actually seems to come from the upath update; they seemed to change the ES5 import style. This would actually allow for cleaner code, so I'm apply the necessary changes now.

IntegerLimit (commented):
Is it possible to add a comment next to each pinned action so we know what version it refers to? (Also, I assume its already done, but I think it would be preferable if all hashes referred to version tags)

Hmm, but they (the comments) wouldn't be updated by dependabot right?

.github/dependabot.yml
```diff
@@ -0,0 +1,37 @@
...
+    directory: "/tools"
+    labels:
+      - "dependencies"
+      - "auto-creation"
+      - "javascript"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
```
IntegerLimit: Let's actually enable updates on npm packages with semver major; as long as its flagged separately (like what you have done with actions) I think its a good idea to keep everything up to date; especially since its sometimes a bit difficult to ensure the buildscripts are up to date with everything.

IntegerLimit (changes requested):

IntegerLimit:
As of #1546, updating npm deps should no longer cause errors.

D-Alessian:
> Is it possible to add a comment next to each pinned action so we know what version it refers to?
> 
> Hmm, but they (the comments) wouldn't be updated by dependabot right?

We could add this via a CI that triggers based on PR that change `.github/workflows`, parses through all the actions pinned to commit SHAs via regex, gets their version tag, and then updates the line above/next to it depending on that version tag.

 > (Also, I assume its already done, but I think it would be preferable if all hashes referred to version tags) 
 
 I'm not quite sure what you mean? Tags don't have a hash themselves, afaik. They only hold the hash of the latest commit from that tag.

D-Alessian:
Ah so actually, dependabot does take care of this as stated [here](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#github-actions)

Syntax: 
`actions/checkout@<commit> #<tag or link>`

D-Alessian:
I just realized my prettier formatted all single quotes to double quotes, do you want this changed back?

IntegerLimit:
> > Is it possible to add a comment next to each pinned action so we know what version it refers to?
> > Hmm, but they (the comments) wouldn't be updated by dependabot right?
> 
> We could add this via a CI that triggers based on PR that change `.github/workflows`, parses through all the actions pinned to commit SHAs via regex, gets their version tag, and then updates the line above/next to it depending on that version tag.
> 
> > (Also, I assume its already done, but I think it would be preferable if all hashes referred to version tags)
> 
> I'm not quite sure what you mean? Tags don't have a hash themselves, afaik. They only hold the hash of the latest commit from that tag.

Sorry, I meant the tag that the hash holds. Like, all current action version hashes correspond to a git commit of a release, not the current state of the main branch or etc.

D-Alessian:
> Sorry, I meant the tag that the hash holds. Like, all current action version hashes correspond to a git commit of a release, not the current state of the main branch or etc.

Oh yeah okay, I used the latest commit of the tags commented next to it indeed. Not the latest from main/master.

IntegerLimit (approved):
Looks good! Thanks for making all the changes tao, and for taking the initiative to make this security change.

PR #1553, opened 2026-04-20
Allow Void World / Lost Cities in Air Collector Circ 1

IntegerLimit:
This PR allows circuit 1 to be used for overworld, void world and lost cities, reducing player confusion.

Circuit 4 for void world / lost cities has been kept for legacy compatibility purposes.

Should it be a hidden recipe?

### DIFF
--- a/overrides/groovy/post/main/general/early/earlygame.groovy
+++ b/overrides/groovy/post/main/general/early/earlygame.groovy
@@ -445,7 +445,7 @@ mods.gregtech.gas_collector.changeByOutput(null, [fluid('air') * 10000])
 
         // LEGACY: Circuit 4 for Void World / Lost Cities
         builder.copyOriginal()
-            .changeCircuitMeta {4 }
+            .changeCircuitMeta { 4 }
             .builder { GasCollectorRecipeBuilder recipe ->
                 // Don't copy properties or add overworld dim to keep legacy behaviour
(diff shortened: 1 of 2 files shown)


D-Alessian (commented):
I think that making the legacy recipe hidden is indeed a better idea.

Other than that, LGTM.

v3ect0rgames (approved):

D-Alessian (changes requested):
After testing it in game, I have a very specific nitpick, the order it appears in JEI is 2,3,1,4. It doesn't feel polished. Nuking all the recipes and building them again would fix this I'd assume, not sure if other fixes are possible.

I also think hiding circ 4 is better, it would confuse players.

IntegerLimit:
Got it.

IntegerLimit:
> After testing it in game, I have a very specific nitpick, the order it appears in JEI is 2,3,1,4. It doesn't feel polished. Nuking all the recipes and building them again would fix this I'd assume, not sure if other fixes are possible.
> 
> I also think hiding circ 4 is better, it would confuse players.

I'm not sure how this didn't appear in my testing, but yeah the scripts would place the new recipe at the back. I'll sort this out.

IntegerLimit:
The plan for this is to use a new Labs helper (coming soon) allowing replacements of recipes without changing recipe order; as this should be generally useful.

PR #1565, opened 2026-05-14
EIO crystal batch craft recipes

v3ect0rgames:
Adds batch craft autoclave recipes for soul binder crystals. Needs discussion over whether to implement this or avoid a greg recipe for eio items.


v3ect0rgames:
This needs discussing and will probably end up in 1.8

PR #1566, opened 2026-05-15
Fully Port CraftTweaker Scripts to GroovyScript

D-Alessian:
Pretty self explanatory. I am making some light changes that I think are better options, will try to list them all here.

### General
- Use ore() instead of item() or metaitem() when applicable. Example `item('frame*')` -> `ore('frameGt*')`

### LazyAE2
- Add PTFE and PBI as optional recipes for recipes using plastic.


D-Alessian:
Mostly used Regex replace for _oreDict.zs, so if you want to review something, probably a good place to spend a lot of time. (When I'm done with it)

PR #1572, opened 2026-05-29
Fix eup2p name in ae2 config to match real name

Splixone:


D-Alessian:
LGTM. Not Tested by myself.

We talked about this in private so I know what this is already. What Splixone explained to me is the config somehow uses the wrong name for EU P2P's toggle, thus making it impossible to disable it, and changing it to this makes it possible to disable it again.

This isn't critical to Nomi-CEu specifically considering EU P2P is enabled, but will avoid us to have to figure this out again in the future if this becomes necessary (e.g. HM changes), and for users who might want to do that.

D-Alessian:
If this is an issue from AE2UEL itself (and not malformed configs in this repo), we should def make a PR there as well. But considering that AE2UEL is pretty stale currently it's a good idea to fix this here anyways

D-Alessian (approved):
Tested, seems like malformed configs on Nomi-CEu's side, not need to report this to AE2

PR #1581, opened 2026-07-01
Slightly Increase Speed of Diamond/Obsidian Furnace

### LINKED ISSUE #1577: Diamond and obsidian furnaces consume incorrect amount of fuel
### Nomi CEu Version

1.7.7

### Launcher

CurseForge

_No response_

### Environment

Singleplayer

### Mode

Expert Mode

### What Happened

When using certain fuels (sticks, sugarcane, saplings, bowls, logs) diamond and obsidian furnaces will consume more fuel than expected;

a 64 stack of sticks can only smelt 30 items not 32, in order to smelt 16 items 34 sticks are required, to smelt 1 item 3 saplings are required, 20 logs only smelt 29 items, 2 logs only smelt 2 items,

all of the smelting ratios provided in this report ignore the obsidian furnace's ability to parallel smelt

other fuels like tiny-coal, coal, charcoal, blaze rods, etc are unaffected by this bug

### Expected Behavior

the diamond and obsidian furnaces should consume the correct amount of fuel expected from the smelting mechanic

### Reproduction Steps

get a diamond or obsidian furnace and try and smelt 1 item using 2 saplings, it will fail to smelt

### Logs

_No response_

### Additional Information

_No response_

IntegerLimit:
With the previous consumption rate multiplier of 8, through the logic of the te, the consumption rate per tick would be fuel burn rates divided by 8. Examples:

`WOOL: 100/8 = 12.5`
`STICK: 100/8 = 12.5`
`LOGS: 300/8 = 37.5`

This is then rounded down. This resulted in certain fuels being less efficient for diamond/obsidian furnaces.

This PR increases the consumption rate multiplier to 10, and the speed accordingly decreasing to 20. This should result in a clean division for essentially all fuels.

(Partially) Fixes #1577

### DIFF
--- a/overrides/config/morefurnaces.cfg
+++ b/overrides/config/morefurnaces.cfg
@@ -3,16 +3,16 @@
 general {
     D:copperFurnaceConsumptionRate=2.0
     I:copperFurnaceSpeed=100
-    D:diamondFurnaceConsumptionRate=8.0
-    I:diamondFurnaceSpeed=25
+    D:diamondFurnaceConsumptionRate=10.0
+    I:diamondFurnaceSpeed=20
     D:goldFurnaceConsumptionRate=5.0
     I:goldFurnaceSpeed=40
     D:ironFurnaceConsumptionRate=1.33
     I:ironFurnaceSpeed=150
     D:netherrackFurnaceConsumptionRate=1.0
     I:netherrackFurnaceSpeed=1800
-    D:obsidianFurnaceConsumptionRate=8.0
-    I:obsidianFurnaceSpeed=25
+    D:obsidianFurnaceConsumptionRate=10.0
+    I:obsidianFurnaceSpeed=20
     D:silverFurnaceConsumptionRate=3.33
     I:silverFurnaceSpeed=60
 }


D-Alessian (commented):
Are we sure there are no fuels with values that are not divisible by 10? If not, LGTM. If there are, the better fix would probably be to round up instead of down through a mixin?

v3ect0rgames:
I just checked, there are blocks that are have fuel values that are not divisible by 10, carpets, wood screws, small and tiny pulp, and coal and coke, small and tiny pulp.

v3ect0rgames (changes requested):
This would not work for the certain fuels which are not divisible by 10

IntegerLimit:
Yes, that is true, but those are non-standard. Furthermore, those fuels already cause issues with the other furnaces. I'm not sure if it is worth fixing, especially since some of them have extremely weird values.

IntegerLimit:
> Are we sure there are no fuels with values that are not divisible by 10? If not, LGTM. If there are, the better fix would probably be to round up instead of down through a mixin?

Rounding up would result in over-efficient fuels, we would still probably have to make it 10 for logs/wood to make it less obvious. There is no way of perfectly fixing this unless we change the logic completely (e.g. change it to x amount of cycles per tick instead of just multipliers)

D-Alessian:
> > Are we sure there are no fuels with values that are not divisible by 10? If not, LGTM. If there are, the better fix would probably be to round up instead of down through a mixin?
> 
> Rounding up would result in over-efficient fuels, we would still probably have to make it 10 for logs/wood to make it less obvious. There is no way of perfectly fixing this unless we change the logic completely (e.g. change it to x amount of cycles per tick instead of just multipliers)

In my opinion, over-efficient is better than under-efficient. Smelting one ore more with a stack sticks or something isn't going to break the balance, but smelting one ore less can be annoying for people who know the ratios it should do.

IntegerLimit:
In my opinion, we would need both fixes; let's have them generally be over-efficient but also apply this PR to avoid too many cases of that.

